PR #26, opened 2017-03-17
Multibuy groundwork

DelusionalLogic:


stregsystem/admin.py
```diff
@@ -21,15 +25,35 @@ def get_room_name(self, obj):
...
     get_room_name.admin_order_field = "room__name"
 
+    def delete_model(self, request, obj):
+        transaction = PayTransaction(obj.price)
+        obj.member.rollback(transaction)
+        obj.member.save()
+        super(SaleAdmin, self).delete_model(request, obj)
+
+    def save_model(self, request, obj, form, change):
+        print(change)
```
fishbacon: Who are we printing for?

DelusionalLogic: I don't know, I just like printing.

It's gone now.

fishbacon:
Jeg savner en fejl når man ikke kan købe et produkt.

fishbacon:
Glem det, jeg kan se der slet ikke er multibuy i den version du vil merge?

DelusionalLogic:
I never removed the len(bought_id) == 1 check, since I wasn't sure if you guys had any other considerations when you disabled it.

As a consequence this doesn't actually enable multibuy, it just does some work to make it clear and testable.

fishbacon:
After many hours of code review I am fine with this going in. There are some bugs that emerge when you enable multibuy but we can handle those when we get to them.

(I looked it up, I did make something similar but less complete two years ago, https://github.com/fishbacon/stregsystem/blob/master/stregsystem/finance/models.py)

This could also lay the groundwork for making the whole accounting thing easier, including making the member balance field obsolete (a dream of mine).

DelusionalLogic:
Yeah, that was kind of the direction I expected this to go eventually. For now it's just a first step.

jones-g:
@fishbacon Please remember as we talked about in at fjubilæum that I remember something about it being a requirement that each member has a balance field. Perhaps @yrke or @mchro can remember this?

mchro:
The balance field is there primarily so the database does not have to calculate it all the time; I don't see the system working efficiently without it. Also, historically the balance field was edited more or less at will, so the sanity check that SUM(transactions) == balance does not hold.

fishbacon:
> Also, historically the balance field was edited more or less at will, so the sanity check that SUM(transactions) == balance does not hold.

Ignoring the monetary aspect, this could be corrected with a balancing transaction created to correct any inconsistencies.

> The balance field is there primarily so the database does not have to calculate it all the time; I don't see the system working efficiently without it. 

Removing the field completely is probably not the right move.

Would the cost to recalculating the balance on e.g. every transaction be too large?

PR #30, opened 2017-03-21
adds activated, formatted price and id to product display in admin interface

VirtualSatai:


stregsystem/models.py
```diff
@@ -198,10 +198,10 @@ class Product(models.Model): # id automatisk...
     deactivate_date = models.DateTimeField(blank=True, null=True)
 
     def activated(self):
```
DelusionalLogic: Maybe this should be in the admin class

stregsystem/admin.py
```diff
@@ -43,9 +43,16 @@ def toggle_active_selected_products(modeladmin, request, queryset):
 class ProductAdmin(admin.ModelAdmin):
     search_fields = ('name', 'price', 'id')
     list_filter = ('deactivate_date', 'price')
-    list_display = ('name', 'price')
+    list_display = ('activated', 'id', 'name', 'get_price_display',)
```
DelusionalLogic: Is this order correct?

VirtualSatai: Yes

DelusionalLogic: Let's ask someone

PR #38, opened 2017-03-22
Turn forms into inputs with javascript action

### LINKED ISSUE #37: Stop restart from speqing csrf token into the query parameter.

fishbacon:
Fixes #37 by using javascript instead of submit actions.

In addition emacs fixed the whitespace.


DelusionalLogic:
I think this is a better way of doing it regardless :+1:

PR #40, opened 2017-03-22
Simple fix for multibuy needing to be atomic.

fishbacon:
A very simple update to make multibuy possible without weird errors.

I have not thoroughly tested this but it should work as it rolls back and throws if you do not have the money.


DelusionalLogic:
I kinda feel like #26 might be a better solution, but then again i wrote it so maybe i shouldn't judge.

I'd like someone else to look at this one.

DelusionalLogic:
This has been solved elsewhere

PR #42, opened 2017-03-23
Added changelog creation script

DelusionalLogic:


fishbacon:
I think making the changelog a mirror of the changes made in git is counterproductive.
We should make it a generally useful document with details on big changes and explanation of features.

VirtualSatai:
Just for some context, the purpose of the script was to use it in sending an email to the treo/forstyrrelsen, when we didn't have a public github page.

Issue #45, opened 2017-04-21
The Kiosk screen are currently served by an external server


VirtualSatai:
At the moment each of the 3 raspberry pi + big screens are served by Jacob Elefsens home page. This is not ideal as he's the only one with access to administrating it. 

We have already asked about it, and he agrees. 

We are considering making it a part of the admin panel of the stregsystem, and limiting the access to a few people, chosen by Forstyrrelsen.

DelusionalLogic:
I think the feature as implemented in #46 is a fantastic addition to the arsenal of advertising opportunities forstyrrelsen has.

I envision moving a lot of the needs currently filled by mails (such as advertising events) could be replaced, or at least greatly supplemented, by these billboard screens.

We'd love input on how this feature can be improved in the future.

jacobelefsen:
Only two of the raspberry pi is served by my domain.  All relevant people has access, both from the fklub and external partners. I move the current solution to the fklub server when I have time.

DelusionalLogic:
#46 is a clean reimplementation of what we believe the php script serving the images is doing.

The reasons for reimplementing it are twofold. First we remove any chance of licensing issues. Secondly, and more importantly, we integrate it into an existing system, which should make maintenance easier in the future (since it wont just be a bunch of disconnected scripts). It might also be possible to do different advertisement based upon other events in the system (such as scheduling advertisements for specific periods, or maybe advertising for foobar every Friday) in the future.

Is it a hard requirement that external partners are able to upload directly? I'd imagine they'd rather drop an email to someone and have us do it for them, but I haven't discussed it with them. if it is a requirement we need to get that into #46.

jones-g:
I think moving this to an app alongside stregsystemet in the django project is a great idea. It is possible but not required that we can integrate with things in the stregsystem like showing the price of certain products live and that sort of deal. 

Regarding external access we already have a solution for this imo. Users of the stregsystemet are disjointed from members. Thus we can just create a user for the external partners that have access to upload on some view we create and Bob's your uncle!

jacobelefsen:
Perhaps you should ask for the requirements to the solution before developing anything.

jones-g:
Well then what are the requirements?

I highly doubt that something has been implemented that cannot be replicated in django by one of the crafty volunteers on this project.

VirtualSatai:
@jacobelefsen Any updates on the requirements?

PR #46, opened 2017-04-21
Kiosk v1

VirtualSatai:
Solves #45


kiosk/static/kiosk/kiosk.css
```diff
@@ -0,0 +1,11 @@
+html, body {
+	height: 100%;
+	margin: 0;
+}
+
+#dummy {
+	min-height: 100%;
```
DelusionalLogic: That's a mistake

requirements.txt
```diff
@@ -1,2 +1,3 @@
-Django==1.10.5
-MySQL-python==1.2.5
+Django==1.10.5
+MySQL-python==1.2.5
+Pillow==4.1.0
```
fishbacon: Is this required?

kiosk/views.py
```diff
@@ -0,0 +1,16 @@
+from django.shortcuts import render
+from django.http import HttpResponse, HttpResponseRedirect
+
+from models import KioskItem
+
+def show_pics(request):
```
fishbacon: `show_pictures` please, it is not much longer and more descriptive.

kiosk/views.py
```diff
@@ -0,0 +1,16 @@
...
+    items = list(KioskItem.objects.all())
+    return render(request, 'allkioskitems.html', locals())
```
fishbacon: I would prefer if we could get away from `locals()` it is convenient but we should probably just build view models or use purpose build dictionaries.

Someone should do a refactor really.

kiosk/admin.py
```diff
@@ -0,0 +1,21 @@
...
+
+class KioskItemAdmin(admin.ModelAdmin):
+	search_fields = ('name', )
+	list_display = ('name', 'active')
+	list_filter = ('active', )
+	list_display_links = ('name', 'active')
+	actions = [set_active_kiosk_item, set_inactive_kiosk_item]
+
+# Register your models here.
+admin.site.register(KioskItem, KioskItemAdmin)
```
fishbacon: #newLineAtTheEndOfFilesPlease

kiosk/templates/kiosk.html
```diff
@@ -0,0 +1,37 @@
...
+
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <title>F-Klubben</title>
+
+    <link href="{% static "kiosk/kiosk.css" %}" rel="stylesheet">
+
+    <script>
```
fishbacon: Move this to a script rather than inline.

kiosk/models.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import unicode_literals
+
+from django.conf import settings
+from django.db import models
+
+# Create your models here.
+class KioskItem(models.Model):
+    name = models.CharField(max_length=100, blank=True, null=True, unique=False)
+    #description = models.CharField(max_length=2000)
+    #uploadedDate = models.DateField(auto_now_add=True)
```
fishbacon: Are these two comments planned or just left out from an earlier version?

kiosk/models.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import unicode_literals
+
+from django.conf import settings
+from django.db import models
+
+# Create your models here.
```
fishbacon: Okay I will.

kiosk/views.py
```diff
@@ -0,0 +1,16 @@
...
+
+def kiosk(request):
+    return render(request, 'kiosk.html', locals())
+
+def find_random_image(request):
+    item = KioskItem.objects.filter(active=True).order_by('?').first()
+    return HttpResponse(item.image.url, content_type="text/plain")
```
fishbacon: I would prefer if the kiosk kept some internal state to ensure that a ring of items is shown rather than a random selection, I am fine with the order being random each round but I think each item should get equal time.

treo/settings.py
```diff
@@ -71,6 +71,7 @@
 
 INSTALLED_APPS = [
     'stregsystem',
+    'kiosk',
```
fishbacon: Good call making it a separate app.

kiosk/views.py
fishbacon: I would also prefer if there was a `item.html` (or `items/image.html`) view for displaying the entire object which can then be embedded in the page, e.g. the request just replaces the body with the body of the resulting item. This might be more of an enhancement though. This would also make "hiding" whatever images are available on the server a lot easier.

requirements.txt
VirtualSatai: Ja, Django bruger Pillow til håndtere ImageFields.

kiosk/views.py
DelusionalLogic: The multiple monitors make having any actual rotation slightly complicated. I imagine a good solution would be sending some id to the client, have them send it back, and have the server consult it for the next id.

fishbacon: > The multiple monitors make having any actual rotation slightly complicated. I imagine a good solution would be sending some id to the client, have them send it back, and have the server consult it for the next id.

No need, a simple cached result being served for a number of seconds works just fine.
I just quickly made a quick and dirty implementation, took like 10 minutes:

    from datetime import timedelta, datetime
    
    items = None
    item = None
    timeout = datetime.now()
    def find_random_image(request):
        global items
        if not items:
            items = list(KioskItem.objects.filter(active=True).order_by('?'))
    
        global item
        global timeout
        now = datetime.now()
        if not item or timeout < now:
            timeout = now + timedelta(seconds=2)
            item = items.pop()
    
        return HttpResponse(item.image.url, content_type="text/plain")

**This is a terrible solution but a version using a cache should work just fine.**

fishbacon: I would also make it more sophisticated, e.g. pushing any new items to the back of the stack.

fishbacon: This also makes it possible to reduce the JS to just `setTimeout(location.reload, 5000)`

fishbacon: Implemented it because I was bored, https://github.com/f-klubben/stregsystemet/tree/feature-kiosk-cached

kiosk/models.py
DelusionalLogic: Planned. If i remember correctly

kiosk/views.py
DelusionalLogic: I agree, but let's do that somewhere else. We almost got caught by this today when we didn't know where something was used when we were refactoring.

kiosk/views.py
DelusionalLogic: My only problem with doing a refresh is that we forfeit our opportunity to do any kind of transition. It's also wasting a bit of bandwidth, but it's negligible.

Using a cache kinda creates a tight coupling between the js and the cache timing. If the timing of the cache is slightly faster than the js, we'll have a problem where we skip some images. If it's slower we'll have some images on there for longer. I personally don't see the importance of the images being synchronized on the monitors, and also don't see how utility in having them ordered. I'm ready to be convinced otherwise, but in this case I don't think the utility outweigh the added complexity.

kiosk/models.py
DelusionalLogic: Good!

kiosk/views.py
fishbacon: My code is just an example, using more complex javascript could be done if transitions is something we really need. I was trying to keep logic out of our views because my experience is that it is not easy to maintain.

The cache timer and the page timer are completely independent, as they are now (although the cache timer is effectively 0). I am afraid I don't see how we would end up skipping pages when using the cache? Please provide an example.

I like the deterministic page display because it makes it easier to wait for something to come around again "When was the FLAN infographic? I know it was right after that Limfjords porter ad!".

Monitors being in sync is really just a side effect of using the cache, if we give each monitor an ID we can give them individual cycles. Although this introduces complexity.

I would argue that the cache gives us options and control and that it is more than worth it to use it.

DelusionalLogic: I might be misunderstanding the cache idea. I'll try and give an example, and you can correct me when I'm wrong:

We have two monitors, A and B. We also have a list, L, consisting of pictures p1, p2 ... pn. The monitors can talk to a Server, S, which knows the list of pictures. A/B will decide how long every picture is displayed, and query the server when a new picture is needed. In this case, they will display every picture for the same time, and also keep timing with each other. They might not tick over at the same time though!

Right now, A can ask S for a picture from L at any time, and it gets a random picture from L. A gives every picture the same time on the monitor. If the monitor runs for an infinite amount of time, then the system will converge on absolute fairness.

If S has a cache, C, things get more complicated. When A queries S two things can happen, either C on S hits or misses. If C hits, then the old response is repeated, but if it misses a new one is generated and added to C. let's just for this example say that C's expiration is set to 5 seconds.

If A picks a timing of 2 seconds per picture, we have a trivial problem. A will actually get a new picture every 6 seconds, because that is the first time it queries S after C has expired.

Add a second client and this turns "catastrophic" (quotes because we're just talking about pictures of advertisements, not anything ACTUALLY serious). Say that every character is half a time unit. ```-``` means wait. for A and B ```|``` and ```\``` means query returning the same and something new respectively. ```+``` means C miss on S, hits won't be denoted since they clutter everything up. P will show the picture loaded into the cache

    B \---|---|---\---|---\---|---|---\---|---\---|---|---\
    S +---------+---------+---------+---------+---------+--
    A --\---|---\---|---|---\---|---\---|---|---\---|---\--
    P 1---------2---------3---------4---------1---------2--

Even though the cache is fair (as can be seen by the even timing between every picture, we can see that this creates unfair behaviour on the monitors. Lets focus on A, although B exhibits the same unfair behaviour.

    P 1---------2---------3---------4---------1---------2--
    A --\---|---\---|---|---\---|---\---|---|---\---|---\--
    E --1-------2-----------3-------4-----------1-------2--

Here I've added E, which shows when every picture was EFFECTIVELY displayed on the monitor. With this view it becomes exceedingly clear that 1 and 3 underrepresented. And the repeating nature shows that they will be underrepresented in the future as well.

That was quite a mouthful.

fishbacon: > Right now, A can ask S for a picture from L at any time, and it gets a random picture from L. A gives every picture the same time on the monitor. If the monitor runs for an infinite amount of time, then the system will converge on absolute fairness.

Sure. I guess that is also what my solution proposes. I would also like to propose that we make the randomization of images happen only once or have them ordered in some way by an administrator. I would like our customers to be able to look at the screen and expect what happens next.

fishbacon: > Cache stuff.

I really do not mind that the cache is not fair. If I said that somewhere I am sorry.

What I want is a deterministic slideshow that our members can watch and understand the order of :)

To prevent the timing thing we could use a keyed cache instead. If a new screen joins it gets the current image of another screen and continues from there on whatever time it has.

I am not convinced that random order is what we want for this feature.

DelusionalLogic: I'm not sure about that either, I just don't believe that a cache is the solution here.

I'd like to propose a different solution. If we give the client a cookie, we can look at that cookie to know what comes next, and give the client a new cookie. By doing so we can store state on the client, which means the client retains control over the timing, but the server controls the ordering. I think that's the perfect split. I see the implementation as giving the client an id for this image, then when we get a request we look at the cookie, find the image after it, and send the client a new cookie identifying this new image.

We can then randomize the image if no cookie if found to allow a bootstrapping of the cookie.

fishbacon: As long as our users can look at the screen and tell what is going on I am for it.

PR #67, opened 2017-09-07
Split product filter tests, and pep8

DelusionalLogic:
This doesn't need to be merged into production, since it only affects
development procedure.


VirtualSatai:
This PR also improves our LOTC (Lines of Test Code) so I like it.

PR #69, opened 2017-09-08
Moved us over to mysqlclient

DelusionalLogic:
mysqlclient is compatible with python 3.5 3.6 and 2.7 in windows

### DIFF
--- a/deploy_requirements.txt
+++ b/deploy_requirements.txt
@@ -1,0 +1,2 @@
+-rrequirements.txt
+mysqlclient==1.3.12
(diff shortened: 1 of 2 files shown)


DelusionalLogic:
This requires manual testing with a local mysql database. Please ignore travis!

VirtualSatai:
Before making the switch doing some kind of local testing would be really helpful.

VirtualSatai:
Since the long term plan is like to go away from MySQL #105, I'll consider this invalid and close it

PR #72, opened 2017-09-12
Added more stats and cleanup of the existing

DelusionalLogic:
The HTML is a bit nicer now, and the actual factual stats are more
plentiful. I don't think they are extremely useful, but they'll be fun
to look at.


DelusionalLogic:
In case you are wondering how it's supposed to look:

2017-09-12-222320_1917x991_scrot

PR #73, opened 2017-09-13
Limited availability products

DelusionalLogic:
Now products can have a limited number. Hopefully we can get this in before flan

Also includes some nice order abstraction to hopefully ease multibuy development!


stregsystem/admin.py
```diff
@@ -112,10 +112,19 @@ def get_price_display(self, obj):
...
+        if (
+            obj.active
+            and (
+                obj.deactivate_date is None
+                or obj.deactivate_date > timezone.now()
+            ) and (
+                obj.quantity is None
+                or obj.quantity > obj.bought
+            )
+        ):
```
fishbacon: If possible change this if to a return and have the different conditions spelled out in variables rather than this.

DelusionalLogic: You're right. Product should obviously have this behavior as a member method. I'll get that done!

fishbacon: Since you are going to change it anyway, what about a descriptive name for instead of `obj`? (eg. product)

DelusionalLogic: Good point, done!

I also took the liberty of refactoring the expression out into some intermediate variables. It's much more clear what the business logic is now.

DelusionalLogic:
> Looks like this also effectively introduces most of multibuy? Okay.

It does, so it's pretty key to get this right. I'm happy you had the time to look it though. Thanks :)

PR #75, opened 2017-09-15
Implemented multibuy

josmad:


stregsystem/templates/stregsystem/error_invalidquickbuy.html
```diff
@@ -0,0 +1,33 @@
...
+{% block message %}
+<div class="center error-container">
+<pre class="the-pre-error">{{ correct }}</pre><pre class="the-error">{{ incorrect }}</pre>
+<pre class="the-error-pointer">
+{{ error_ptr }}
+{{ error_msg }}
+</pre>
+</div>
+<div class="ebnf-container">
+<div class="center">HINT! Se følgende EBNF:</div>
```
fishbacon: This makes me happy.

stregsystem/views.py
```diff
@@ -44,31 +42,32 @@ def index(request, room_id):
...
+    # Extract username and product ids
+    try:
+        username, bought_ids = QuickbuyParser.parse(buy_string)
+    except QuickBuyError as err:
+        values = {
+            'correct': err.parsed_part,
+            'incorrect': err.failed_part,
+            'error_ptr': '~' * (len(err.parsed_part)) + '^',
+            'error_msg': ' ' * (len(err.parsed_part) - 4) + 'Fejl her'}
+        return render(request, 'stregsystem/error_invalidquickbuy.html', values)
```
fishbacon: Feature request, an error page for when a multibuy includes a product we don't sell, e.g. `tester 1 99`. It could even be this page.

stregsystem/utils.py
```diff
@@ -1,7 +1,78 @@
...
+        self.parsed_part = parsed_part
+        self.failed_part = failed_part
+
+class QuickBuyParseError(Exception):
+    pass
+
+class QuickbuyParser(object):
+    _item_matcher = regex.compile(
+        r'(?V1)(?P<productId>\d+)(?::(?P<count>\d+))?$'
+    )
```
fishbacon: I never think an agreement was reached on how multibuy should be represented, this is my favourite but I imagine others might want something different. (essentially some suggested a `<product> <count>` idea, maybe that should be an optional special syntax? FEATURE REQUEST!

fishbacon:
I also appreciate the branch name.

stregsystem/utils.py
```diff
@@ -1,7 +1,78 @@
...
+
+class QuickBuyError(Exception):
+    def __init__(self, parsed_part, failed_part):
+        self.parsed_part = parsed_part
+        self.failed_part = failed_part
+
+class QuickBuyParseError(Exception):
+    pass
+
+class QuickbuyParser(object):
```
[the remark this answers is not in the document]
josmad: Agreed

stregsystem/utils.py
josmad: Personally I do not like something like `<product> <count>` as it requires additionaly syntax to support multiple products in a single buy.
I am very much open to input on the syntax, I made a decision with some inputs from @VirtualSatai and @DelusionalLogic

stregsystem/views.py
josmad: We could do this, it could be useful, but it would change the behaviour of the stregsystem. Pre-multibuy behaviour is to show the usermenu when a product does not exist.
https://github.com/f-klubben/stregsystemet/blob/54b4fa9cb6683c5032846f9c0351d52f6f08e8da/stregsystem/views.py#L86-L87

stregsystem/utils.py
fishbacon: It really comes down to what our customers would prefer. I have no preference.

DelusionalLogic:
If you fix the requirements I'll just merge it in!

PR #76, opened 2017-09-17
Feature more tests

DelusionalLogic:
I've based this on multibuy, because we are almost 100% gonna merge that, but it should be changed before we actually merge.


DelusionalLogic:
WE FINALLY MADE IT TO 80%! Yellow, here we are.

PR #79, opened 2017-09-18
[WIP] Added a CONTRIBUTING file

fishbacon:
This needs work. I feel like the example is not that great especially
since it does not match the way we are currently looking at
multi-buy.

I am thinking maybe make it a named products example instead.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,11 @@ If you want to contribute code you should fork the project.
 The project strives to be Python 2 and 3 compliant, if you are writing a feature which will only work in Python 3 please note this clearly and explain why.
 
 ### Coding standards
-Travis runs `flake8` on every pull request, so adhere to what flake8 tells you.
+We try to adhere to `flake8` whenever possible. Mostly to avoid any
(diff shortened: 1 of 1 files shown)


fishbacon:
This is in response to #3.

CONTRIBUTING.md
```diff
@@ -0,0 +1,46 @@
...
+> Customers often want to buy several beers in a row and the current workflow of typing in your username and then using to mouse to click on the product can be sometimes frustrating.
+> I suggest the ability to enter `<username> <product id 1> <number of products sought>` thus making it possible to buy several products in one purchase.
+
+## Patches welcome!
+If you want to contribute code you should fork the project.
+
+The project strives to be Python 2 and 3 compliant, if you are writing a feature which will only work in Python 3 please note this clearly and explain why.
+
+### Coding standards
+Travis runs `flake8` on every pull request, so adhere to what flake8 tells you.
```
DelusionalLogic: Is unfortunately false. Currently Travis does not run flake8. I personally handle that in a pre-commit hook (which flake8 can install).

Issue #78, opened 2017-09-18
Add category to products 


Rene4100:
For now, when new products are added to the Stregsystem you(FIT) have to manually the rankings found in report/ranks/ this could be done easier by adding categories to the products.

I know it might be a lot of work to update the old products with a category, but I am willing to help out with it.

VirtualSatai:
I think this is doable in a reasonable way, without much hassle. I image that some items would be part of multiple categories. (Category <-> Product is a many-to-many) relation. I will take a look at it soon (tm).

Can you imagine any other places, than the sales ranks, where this could be relevant?

fishbacon:
Started over here: https://github.com/f-klubben/stregsystemet/tree/feature/product-categories

VirtualSatai:
Since this is merged, and the remaining points on the list is enhancements, I'll close the issue.

PR #80, opened 2017-09-18
Feature product categories

fishbacon:
Working on #78.

- [x] Create category  
- [x] Merge commit because I branched from my local next
- [x] Add categories to admin page
- [x] Change multiple migrations to single migration for pull request
- [x] Add to ranking pages
- [x] Tests!

Nice to have features, possible enhancements for after merge:
- [ ] Link to the category from products listing
- [ ] Instantly add products to new category (product listing on edit category page)
- [ ] Add fields for start and end time of the ranking


fishbacon:
> Maybe this isn't super important since we're really in going to use it in the admin interface, but just getting it out of the code is worth it to me.

Down the line categories could do more but I agree that the feature is currently quality of life for our volunteers.

I think using this feature to improve stregreport would be great.

fishbacon:
With this commit we kinda fulfil the whole issue and I am probably going to recommend merging now.

I will be creating issues for the rest of the things in the list in the top comment.

treo/settings.py
```diff
@@ -183,7 +183,7 @@
 
 USE_L10N = True
 
-USE_TZ = True
+USE_TZ = False
```
DelusionalLogic: I'm not sure what that's doing in here.

fishbacon: Dunno, probably slipped up. This was to avoid the errors related to #84. Will revert it.

josmad:
I cleaned up and squashed the commits and rebased it on top of next in preparation to merge

PR #87, opened 2017-09-22
Correct Salgsrapportering

### LINKED ISSUE #85: Salgsraporteringer does not work
Currently Salgsraporteringer does not work, when I try to use it

VirtualSatai:
This fixes #85 

The old method checked the type of the input, this was given as unicode, but only worked if it had type string. Now we just try, if we fail, we'll report the error to the user.

### DIFF
--- a/stregsystem/templates/admin/stregsystem/report/error_invalidsalefetch.html
+++ b/stregsystem/templates/admin/stregsystem/report/error_invalidsalefetch.html
@@ -1,16 +1,24 @@
 ﻿{% extends "admin/base_site.html" %}
 
-{% block title %}Stats for {{year}}{% endblock %}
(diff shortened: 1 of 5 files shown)


stregsystem/templates/admin/stregsystem/report/error_invalidsalefetch.html
```diff
@@ -1,16 +1,24 @@
...
\ No newline at end of file
+    <center>
+        <div id="headline" style="color: red;">
+            <h3>Der kunne ikke hentes produktinformation for de angivne produkter i de angivne datoer.</h3>
+        </div>
+        <br>
+        Du fik fejlen: <b> {{ error }} </b>
+        <br>
+        <br>
+        Hvis du mener at <i>det er for galt</i>, anbefaler F-Klubbens IT, at du venter, indtil nogen gør noget ved problemet.<br/>
```
fishbacon: Patches welcome, link til github?

DelusionalLogic:
We should probably have more tests for instances where there are actually an sales.

This'll do for now though!

PR #88, opened 2017-09-23
Partial autofocus, at least on the menu.html page

DelusionalLogic:
I guess you can access the "restart" link with Alt+q (Alt+Shift+q in
firefox). Maybe this is sufficient?

### DIFF
--- a/autofocus/admin.py
+++ b/autofocus/admin.py
@@ -1,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
--- a/autofocus/apps.py
+++ b/autofocus/apps.py
@@ -1,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class AutofocusConfig(AppConfig):
+    name = 'autofocus'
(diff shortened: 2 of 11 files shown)


hrjakobsen:
Is this not a bit of an overkill for focusing a single element? Would it not be enough to make sure that the link is the first element in the tab-ordering?
If we change the link to `<a tabindex="1" href="/{{room.id}}/user/{{member.id}}">Bruger Info</a>`, it will be the first element focused when pressing tab.

DelusionalLogic:
It's a lot of "code" for very little functionality now. I argue it isn't a problem, since none of the code is actually active nor requires support.

It's also important to note that this feature is a direct response to a request by a member. What the member wanted was to be able to just always click enter to return to the main page. I tried to make this happen, but I'm having trouble doing so with the restart button since it's in the base package.

There's some stuff we can do, but this is a first step.

kimsey0:
If you're going to merge this sometime, may I suggest changing from using the id `autofocus` to using the same `autofocus` attribute as for input elements? `document.getElementById("autofocus")` can just be changed to `document.querySelector("[autofocus]")`.

hrjakobsen:
I think we have moved with other solutions than this, and as such I will close this PR.

PR #89, opened 2017-09-23
Makes multiselect a bit nicer, and change table

DelusionalLogic:
The table now shows the individual sales for every category, as well as
the total across categories.

The code is really messy, but at least it's done in the database, so
yay?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ Coverage==4.4.1
 pytz==2017.2
 regex==2017.07.28
 git+https://github.com/spulec/freezegun.git@385a78d004b0566c6cbf98c304fb89a57bad0005
+Django-Select2==5.11.1
--- a/stregreport/forms.py
+++ b/stregreport/forms.py
@@ -1,9 +1,15 @@
 from django import forms
+from django_select2 import forms as s2forms
 
-from stregsystem.models import (Category, Product)
+from stregsystem.models import Category
+
 
 class CategoryReportForm(forms.Form):
     categories = forms.ModelMultipleChoiceField(
         required=True,
+        label="Category",
+        widget=s2forms.ModelSelect2MultipleWidget(
+            search_fields=["name__icontains"],
+        ),
         queryset=Category.objects.all()
     )
--- a/stregreport/views.py
+++ b/stregreport/views.py
@@ -1,16 +1,14 @@
-import collections
 import datetime
 from functools import reduce
 
 from django.contrib.admin.views.decorators import staff_member_required
-from django.db.models import Count, Q, Sum
+from django.db.models import Count, Q, Sum, F
 from django.db.models.functions import TruncDay
 from django.http import JsonResponse
 from django.shortcuts import render
 from django.utils import timezone
 
 from stregsystem.models import (
-    Category,
     Member,
     Product,
     Sale,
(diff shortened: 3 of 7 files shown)


DelusionalLogic:
I don't know how I fucked the requirements.txt file up 3 TIMES. (I pushed, then amended 2 times, each time failing to include the requirements.txt file.)

I'm trying to incrementally "fix" the layout of the admin panels as we go along. Everything I've touched looks similar (daily stats has the same no grid style as this). The others aren't updated because I haven't touched them. I'm hesitant to spend any time on them, since some from treo'en have called them near useless.

Select2 could be included in the ModelAdmins. It shouldn't be that hard, since the library has all the widgets you need.

For the last point I don't think either is really all that useful. I can't think of a case where it would really matter who bought the most "alcohol AND caffeine" or "alcohol OR caffeine". For a more interesting/useful use of categories I think we should look into listing the most popular category, or even the most popular products in a specific category. I could imagine Treo'en wanting to know which energy drink is most popular, or what beer is the most popular, and by how much.

I don't know how much of the it you can actually factor out. That's the pain of constructing queries.

fishbacon:
Should we consider changing the page to be top X for each category, so a table for each like in the yearly ranking page?

I take your point of updating as we go and concede on that point.

DelusionalLogic:
> Should we consider changing the page to be top X for each category, so a table for each like in the yearly ranking page?

I'm worried about the amount of queries a page like that would create. The visual clutter of that many tables would also make it hard to read. I don't know how often you'd want to see all the tables at once.

I think we should just limit this page to a single term instead of allowing multiple, that seems like an easy way to skirt all the issues with the current design.

If we do want all information at once I think we should consider making it a single table, with multiple columns, that you can sort by clicking on the heading. Since we're already using C3 for the graph in the dashboard D3 seems like a nice fit.

fishbacon:
> I'm worried about the amount of queries a page like that would create. The visual clutter of that many tables would also make it hard to read. I don't know how often you'd want to see all the tables at once.

The yearly report is the use case, it makes it easy to quickly generate that report. 

Why is the number of queries an issue?

> If we do want all information at once I think we should consider making it a single table, with multiple columns, that you can sort by clicking on the heading.

In essence this is what I want and would work just fine.

DelusionalLogic:
> Why is the number of queries an issue?

I guess you could do it all in a single query. That was just me being stupid.

> In essence this is what I want and would work just fine.

I'll have to learn about D3js then!

VirtualSatai:
Worrying about the number of queries seems like premature optimization to me.

DelusionalLogic:
The amount of queries I was thinking it was going to involve would have been important to consider. It doesn't matter though, since it can be accomplished with a single query.

fishbacon:
> Which yearly report? I didn't know we did one.

We do a yearly ranking of who bought the most of a certain type of item. That is why the top-list is xmas party to xmas party :)

I just remembered that what you are proposing would not give the same overview since it does not afford the same general overview, e.g. if I want to see both caffeine and alcohol I am not able to since they are sorted very differently (the same people are not buying a lot of both is the assumption).

Multiple tables, one per category is the best way of solving the issue in my opinion.

I approve this change for now and we should consider a different approach later.

DelusionalLogic:
I agree that we should find a better approach. I'll just merge it in since we have a meeting monday and i want to get this in this merge window.

PR #92, opened 2017-09-25
Ballmer peak!

VirtualSatai:


DelusionalLogic:
That's a strange travis problem!

VirtualSatai:
Yep, [I'll try something](http://s2.quickmeme.com/img/4a/4a102f0529cd0acddfa88a8dea431001d8e671f9def8ecaa6992bc455cee56f4.jpg)

stregsystem/booze.py
```diff
@@ -66,16 +66,31 @@ def alcohol_bac_timeline(gender, weight, now, alcohol_timeline):
         current += alcohol_bac_increase(gender, weight, ml)
 
         # last_time can never be None after the first iteration
-        assert(last_time is not None)
+        assert (last_time is not None)
 
     # Since we return if the list is empty we must have some last time
-    assert(last_time is not None)
+    assert (last_time is not None)
```
DelusionalLogic: Alright, but lets remove the parens too then.

stregsystem/booze.py
```diff
@@ -66,16 +66,31 @@ def alcohol_bac_timeline(gender, weight, now, alcohol_timeline):
...
     return current
+
+
+# Ballmer peak: 1.337 +/- 0.05
+BALLMER_PEAK_MEAN = 1.337
+BALLMER_PEAK_LOWER_LIMIT = BALLMER_PEAK_MEAN - 0.05
+BALLMER_PEAK_UPPER_LIMIT = BALLMER_PEAK_MEAN + 0.05
+
+
+def time_to_ballmer_peak_expiry(bac):
```
DelusionalLogic: I'd say this should be generalized to some other method taking a current and future bac, and returning the timedelta to reach that future bac.

stregsystem/booze.py
```diff
@@ -66,16 +66,31 @@ def alcohol_bac_timeline(gender, weight, now, alcohol_timeline):
...
     if current < 0:
         current = 0
 
     return current
+
+
+# Ballmer peak: 1.337 +/- 0.05
+BALLMER_PEAK_MEAN = 1.337
+BALLMER_PEAK_LOWER_LIMIT = BALLMER_PEAK_MEAN - 0.05
+BALLMER_PEAK_UPPER_LIMIT = BALLMER_PEAK_MEAN + 0.05
```
DelusionalLogic: I like the globals, but let's only use them in this file.

stregsystem/views.py
```diff
@@ -105,6 +106,11 @@ def quicksale(request, room, member, bought_ids):
         return render(request, 'stregsystem/error_stregforbud.html', locals())
 
     promille = member.calculate_alcohol_promille()
+    is_ballmer_peaking = BALLMER_PEAK_LOWER_LIMIT < promille < BALLMER_PEAK_UPPER_LIMIT
```
DelusionalLogic: Maybe extract most of this (at least the is_ballmer_peaking) into a function in booze.

stregsystem/booze.py
VirtualSatai: I did consider that, but didn't want to bother with doing regression on previous data or something of that kind.

stregsystem/views.py
VirtualSatai: Sure, why not

stregsystem/booze.py
DelusionalLogic: When would you need that? Wouldn't it just be exactly the same as you have now?

PR #94, opened 2017-09-27
More enforceable limits

DelusionalLogic:
This patch gets around the transaction on by issue by making a limit
a quantity and a start date. When we have to find out how many we have
already sold we then just tally up the number of sales if this product
since the start date. The query to do so is slightly complicated, but it
works as it stands.

This is more robust and less error prone than the method already merged.

It does have the potential to launch a lot of queries, we will have to
monitor this when we deploy!


DelusionalLogic:
The rebase was a bit of a bitch, but i think i'm ready for a test. We need this merged before the window closes monday.

PR #95, opened 2017-09-29
Increased max length of charfields in model and db

VirtualSatai:
SQLite doesn't check this, right now some of our testdata is incompatible with MySQL since it does enforce the max_length. Also there are no real reasons for these arbitrary limitations.


VirtualSatai:
Also migrations are technically tested, and therefore increased our test coverage. I'll be looking into this.

fishbacon:
I want to see novels published trough stregsystem products.

VirtualSatai:
If nothing else, now it's easier to embed images in the name of products 🤔

fishbacon:
well injecting html is still possible so...

VirtualSatai:
It's a feature!

fishbacon:
> It's a feature!

I am pretty sure we would get some serious pushback from the treo if we suggested not having this feature.

PR #97, opened 2017-10-02
Razzia Wizzard

### LINKED ISSUE #58: Razzia wizard
Someone suggested that the razzia feature was simply too much to bare and that is why we should extend it severely.

I suggest making a razzia wizard, giving the persons who are wanting to check something with regards to members and purchases ample ability to do so.
My suggestion is to create a page where the TREO can enter several products which they want to check, and a title for their razzia. 
More features like max number of re-entries and an error/success message template would also make the experience much nicer. 
A nice-to have would be a model for razzias so that they can be reused, e.g. bread razzia, but I see it as optional.
I imagine the whole thing can be build in a few hours.

VirtualSatai:
This closes #58 

The wizzard itself
billede

The view before entering anything
billede

The view with one successful result
billede

With one success and one fail
billede

Only one fail
billede

Invalid user
billede

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-freezegun
 Django==1.10.5
 Pillow==4.1.0
 Coverage==4.4.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+freezegun
 Django==1.10.5
 Pillow==4.1.0
 Coverage==4.4.1
--- a/stregreport/views.py
+++ b/stregreport/views.py
@@ -4,8 +4,10 @@ from functools import reduce
 from django.contrib.admin.views.decorators import staff_member_required
 from django.db.models import Count, Q, Sum, F
 from django.db.models.functions import TruncDay
+from django.forms import extras, fields
 from django.http import JsonResponse
-from django.shortcuts import render
+from django.shortcuts import render, redirect
+from django.urls import reverse
 from django.utils import timezone
 
 from stregsystem.models import (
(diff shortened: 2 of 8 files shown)


VirtualSatai:
This closes #58 btw.

requirements.txt
```diff
@@ -1,3 +1,4 @@
+freezegun
```
DelusionalLogic: Freezegun is already in

stregreport/views.py
```diff
@@ -59,6 +61,97 @@ def bread_view(request, queryname):
...
+    for res in result:
+        product_dict[res.product.name] += 1
+
+    return product_dict
+
+
+def razzia_view(request):
+    default_start = datetime.datetime.now() - datetime.timedelta(days=-180)
+    default_end = datetime.datetime.now()
+    start = request.GET.get('start', "{}-{}-{}".format(default_start.year, default_start.month, default_start.day))
```
DelusionalLogic: Wutface. isotime does what you want here

stregreport/views.py
```diff
@@ -59,6 +61,97 @@ def bread_view(request, queryname):
...
+        return render(request, 'admin/stregsystem/razzia/wizard_view.html',
+                      {
+                          'start': start,
+                          'end': end,
+                          'products': products,
+                          'username': username,
+                          'razzia_title': title}
+                      )
+
+    start_date = datetime.datetime.strptime(start, "%Y-%m-%d").date()
```
DelusionalLogic: This is what you want: django.utils.dateparse

stregreport/views.py
```diff
@@ -59,6 +61,97 @@ def bread_view(request, queryname):
...
 
 
+def _sales_to_user_in_period(username, start_date, end_date, product_list, product_dict):
+    result = Sale.objects.prefetch_related('member').filter(
+        member__username__iexact=username,
+        product__in=product_list,
+        timestamp__gte=start_date,
+        timestamp__lte=end_date)
+    for res in result:
+        product_dict[res.product.name] += 1
```
DelusionalLogic: This could be done in pure sql.

stregsystem/templates/admin/stregsystem/razzia/bread.html
```diff
@@ -29,7 +29,7 @@
 			{% if member %}
 				<div class="status">
 					<div class="fa fa-check-circle sucess" aria-hidden="true"></div></div>
-				{{member.firstname}} {{member.lastname}} ({{member.username}}) 
+				{{member.firstname}} {{member.lastname}} ({{member.username}})
```
DelusionalLogic: This seems unrelated

PR #107, opened 2017-10-06
Added a hint to use multibuy when people are not

VirtualSatai:


josmad:
What about sales through menusale?
Consider extracting the check to a separate function.

PR #101, opened 2017-10-02
Add spooky bats to stregsystemet. One bat is added each day

mathiassmichno:
Must be included in production quickly to spook people for Halloween 🎃

PR #108, opened 2017-10-09
Add advent candle in december

hrjakobsen:
Inspired by #101 and a suggestion by a fember, this pull request adds an advent candle to stregsystemet in December. I'm not sure what the size of the candle should with the new screen and all, so I just picked something that looked alright on my 1920x1080 screen.

Candle on December 1st

Candle on December 12th

Candle on December 24th (evening)

### DIFF
--- a/stregsystem/static/stregsystem/stregsystem.css
+++ b/stregsystem/static/stregsystem/stregsystem.css
@@ -69,3 +69,16 @@ table h2, table h3 {
 .ebnf-container {
     margin-top: 20px;
 }
+
+.day_hidden {
+  display: none !important;
+  fill: none !important;
+}
+
+#advent_candle {
+  position: fixed;
+  bottom: 0;
+  right: 0;
+  max-height: 75%;
+  max-width: 400px;
+}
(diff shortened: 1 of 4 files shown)


hrjakobsen:
I originally implemented the flame as a css transition, but removed it as I feared that the pi couldn't handle it. I can easily change the flame.

Something like this would be really cool: https://codepen.io/2xsamurai/pen/EKpYMg

Do you think the Pi can handle it?

VirtualSatai:
billede

hrjakobsen:
I changed the flame to the one in the codepen above.

stregsystem/static/stregsystem/stregsystem.css
```diff
@@ -82,3 +82,35 @@ table h2, table h3 {
   max-height: 75%;
   max-width: 400px;
 }
+
+.flame-main {
+  /*animation-name: flameWobble;*/
```
hrjakobsen: If this line is uncommented, there will be a slow fire animation. I think the Pi can handle it, but it is disabled by default.

DelusionalLogic:
It needs to be tested on a PI, but otherwise I think it's pretty nice.

We should also consider how this will look on other screens and other aspect ratios. I guess you could always ```display: none``` in a mediatag if it's a problem.

PR #110, opened 2017-10-12
Disable autocomplete in index template

anton-christensen:
This pull request disables autocomplete in chrome. If it's not disabled, chrome will list previous inputs. A privacy concern for those whose drinking habits are considered unhealthy by the general public.

### DIFF
--- a/stregsystem/templates/stregsystem/index.html
+++ b/stregsystem/templates/stregsystem/index.html
@@ -28,7 +28,7 @@
 
 
 <center>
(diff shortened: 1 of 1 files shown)


VirtualSatai:
I think previously this have just been disabled in the individual browsers running the stregsystem, but it should be part of the code as well, merged!

VirtualSatai:
Also, @anton-christensen Thanks for contributing!!!

Issue #112, opened 2017-10-14
Update multi-buy hint to show an example


Casperbart:
The current hint when a user buys the same thing multiple times in a row says something along the line of “psst... multi-buy is enabled”. 
This message is not very helpful for people who doesn’t know what multi-buy or quick buy is.

A suggestion to improve this is to show a context aware hint e.g. “Multi-buy is enabled. You could have typed {user} {item}:{amount}”. Where amount is the times the person bought the same item, which should be updated every time an item is bought.

DelusionalLogic:
Queued for next release

PR #115, opened 2017-10-16
Add context aware multibuy hint

Flarenityy:
Addresses the issue put forward by Casper in #112 to update multi-buy hint to show an example

### DIFF
--- a/stregsystem/views.py
+++ b/stregsystem/views.py
@@ -87,18 +87,32 @@ def sale(request, room_id):
 
 
 def _multibuy_hint(now, member):
-    # Get a timestamp to fetch sales for the member since.
+    # Get a timestamp to fetch sales for the member for the last 60 sec
     earliest_recent_purchase = now - datetime.timedelta(seconds=60)
-    # Count the sales since
-    number_of_recent_distinct_purchases = (
+    # get the sales with this timestamp
+    recent_purchases = (
         Sale.objects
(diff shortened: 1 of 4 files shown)


VirtualSatai:
Casper and I spoke about this on the fyttetur and I agree with his suggested change :) I'll try this out later and test if it's a-ok. But looks good 👍

VirtualSatai:
I think it's good, but it ends up taking up too much space. Maybe we should change til hint to be bold? or h2/h3 ?

Flarenityy:
@VirtualSatai  I tried the different options, and my personal reference is to change hints to be bold - I've done this

stregsystem/views.py
```diff
@@ -87,18 +87,32 @@ def sale(request, room_id):
...
+                sale_dict[str(sale.product.id)] = 1
+            else:
+                sale_dict[str(sale.product.id)] = sale_dict[str(sale.product.id)] + 1
+        sale_hints = [member.username]
+        for key in sale_dict:
+            if sale_dict[key] > 1:
+                sale_hints.append("{}:{}".format(key, sale_dict[key]))
+            else:
+                sale_hints.append(key)
+        return (True, ' '.join(sale_hints))
```
VirtualSatai: Returning True/False now is not needed since: `bool("")` is False and any other string is True, one can just use that for the check later. Though it doesn't really matter, and being explicit might just be fine.

PR #116, opened 2017-10-17
Made all datetimes timezone aware

VirtualSatai:
This will correct all the annoying warnings we get. 

We should always use `from django.utils import timezone` and `timezone.now()` to get a timestamp from now on please.

### DIFF
--- a/stregreport/views.py
+++ b/stregreport/views.py
@@ -74,8 +74,8 @@ def _sales_to_user_in_period(username, start_date, end_date, product_list, produ
 
 
 def razzia_view(request):
-    default_start = datetime.date.today() - datetime.timedelta(days=-180)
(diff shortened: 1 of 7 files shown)


DelusionalLogic:
I saw you work on this, at it seemed fine! Have you tested if anything strange happens in the DB?

VirtualSatai:
@DelusionalLogic Not directly, will test tomorrow.

VirtualSatai:
Looking at the changes again, they are all for queries for the database, and not actual data, so I don't really see what would be different wrt. to it.

VirtualSatai:
Rebased with origin/next

DelusionalLogic:
That's a good point!

PR #130, opened 2017-11-10
Fix bug with images not being scaled properly

hrjakobsen:
If the images in the kiosk system are too large, they are not auto-scaled to fit the window size. This PR fixes this.

### DIFF
--- a/kiosk/static/kiosk/kiosk.css
+++ b/kiosk/static/kiosk/kiosk.css
@@ -10,7 +10,11 @@ html, body {
 }
 
 .dummy {
+	background-color: black;
 	position:absolute;
+    background-size: contain;
+    background-repeat: no-repeat;
+    background-position: center; 
 	top: 0;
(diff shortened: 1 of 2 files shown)


kiosk/static/kiosk/kiosk.css
```diff
@@ -1,26 +1,29 @@
...
-	left: 0;
-	right: 0;
-	bottom: 0;
-	transition: opacity 0.2s 0s ease;
-	opacity: 0;
+    background-color: black;
+    position:absolute;
+    background-size: contain;
+    background-repeat: no-repeat;
+    background-position: center;
```
DelusionalLogic: I'd personally prefer the ```top```, ```left```, ```right```, and ```bottom``` tags to be kept together with the ```position:absolute``` (since the position attr influences how they are interpreted)

It's super minor though, so we can just fix it the next time we touch the file.

Issue #131, opened 2017-11-15
Improve the Kiosk app


VirtualSatai:
The two screens we use for advertisements are served by the Kiosk application in this repository. This could be expanded to other functions that simple rotating pictures every 10th second. 

Maybe one could show specific advertisements at specific points in time (thursdays fkult, friday foobar etc). 

Maybe one could make a countdown for events, such as fredagsfranskbrød every wednesday before 10:00, or count down to F-Julestue etc. 

The possibilities are near-endless here!

PR #134, opened 2017-11-17
Websites in the kiosk rotation

hrjakobsen:
Adds the ability to show webpages on the kiosk screens with a new kiosk website item. Can be used to implement countdowns and live content as described in #131.

This will not merge current kiosk data to the new format; something to be kept in mind when/if merged into master.


kiosk/views.py
```diff
@@ -21,7 +21,8 @@ def find_random_image(request):
 
     response_data = {
         "id": item.id,
-        "url": item.image.url,
+        "content": item.getContent(),
+        "type": item.__class__.__name__
```
DelusionalLogic: I think using the class name (and especially magic methods) might be a bit fragile. I think it would be better if you made a polymorphic method that returned some static string (like jstype or something)

That way we don't need to worry about changing the class name in the js!

hrjakobsen:
I have created templates on the server instead of creating and deleting objects in the javascript based on a suggestion from @fishbacon. I have also made changes according the reviews by @DelusionalLogic and @VirtualSatai.

PR #140, opened 2018-02-10
Batch insertion of payments

DelusionalLogic:
How about this?

image

Testing this is pretty difficult (it's mostly js), so we should probably do some smoke testing.

### DIFF
--- a/stregsystem/templates/admin/stregsystem/batch_payment.html
+++ b/stregsystem/templates/admin/stregsystem/batch_payment.html
@@ -19,8 +19,8 @@
 		{{ formset.management_form }}
 		<table>
 			<tbody>
-				{% for form in formset %}
-				<tr>
+				{% with formset.empty_form as form %}
+				<tr id="formTemplate" style="display: none">
 					{{ form.id }}
 					<td>{{ form.member }}{{ form.member.errors }}</td>
 					<td>{{ form.amount }}{{ form.amount.errors }}</td>
@@ -28,7 +28,7 @@
 						<td>{{ form.DELETE }}</td>
 					{% endif %}
 				</tr>
-				{% endfor %}
+				{% endwith %}
 			</tbody>
 		</table>
 		<input type="submit" value="Submit" />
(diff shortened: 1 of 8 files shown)


VirtualSatai:
It should probably be clear that the field on the right is in hundreds of a stregdollar.

DelusionalLogic:
It's in stregøre actually

hrjakobsen:
With the latest commit it is no longer possible to reach the delete button by pressing tab. This is due to a request from the treasurer.

stregsystem/templates/admin/stregsystem/batch_payment_done.html
```diff
@@ -0,0 +1,14 @@
+{% extends "admin/base_site.html" %}
+
+{% block title %}Brødrazzia{% endblock %}
```
hrjakobsen: This is not the Brødrazzia page

stregsystem/templates/admin/stregsystem/batch_payment.html
```diff
@@ -0,0 +1,68 @@
+{% extends "admin/base_site.html" %}
+{% load static %}
+
+{% block title %}Brødrazzia{% endblock %}
```
hrjakobsen: This is not Brødrazzia

stregsystem/views.py
```diff
@@ -210,3 +217,28 @@ def menu_sale(request, room_id, member_id, product_id=None):
     # Refresh member, to get new amount
     member = Member.objects.get(pk=member_id, active=True)
     return usermenu(request, room, member, product, from_sale=True)
+
+
+@staff_member_required()
+def batch_payment(request):
+    PaymentFormSet = forms.modelformset_factory(
+        Payment,
+        fields=("member", "amount"),
```
hrjakobsen: This allows for negative amounts, which is not intended (right?)

stregsystem/templates/admin/stregsystem/batch_payment_done.html
```diff
@@ -0,0 +1,14 @@
...
+{% block breadcrumbs %}<div class="breadcrumbs"><a href="../../../">Hjem</a>&nbsp;&rsaquo;&nbsp;<a href="../../">Stregsystem</a>&nbsp;&rsaquo;&nbsp;<a>Razzia</a>&nbsp;&rsaquo;&nbsp;Bread</div>{% endblock %}
```
hrjakobsen: Wrong breadcrumbs as well

stregsystem/templates/admin/stregsystem/batch_payment.html
```diff
@@ -0,0 +1,68 @@
...
+{% block breadcrumbs %}<div class="breadcrumbs"><a href="../../../">Hjem</a>&nbsp;&rsaquo;&nbsp;<a href="../../">Stregsystem</a>&nbsp;&rsaquo;&nbsp;<a>Razzia</a>&nbsp;&rsaquo;&nbsp;Bread</div>{% endblock %}
```
hrjakobsen: Wrong breadcrumbs

hrjakobsen:
When multiple payments for the same user are submitted, both payments are saved but the user has a wrong balance.

DelusionalLogic:
I think it's because of the way transactions are handled. I completely forgot that I have to actually add the amount to the users balance. Woops

stregsystem/views.py
DelusionalLogic: It is

DelusionalLogic:
It should be fixed now. Please test!

PR #141, opened 2018-02-15
Added splashscreen to kiosk. 

### LINKED ISSUE #135: Kiosk splash screen
I think it would be nice to have a "default" picture for the kiosk to load. Something that's just static for the first 10 seconds while we wait for the first image to load.

I've made something simple that combines the AAU style with our name, just to kinda keep some semblance of brand identity.

screen

hrjakobsen:
Closes #135


DelusionalLogic:
In think this looks good!

Issue #137, opened 2017-12-02
Upgrade to Django 2.0 and Python 3


VirtualSatai:
Django 2.0 is now out: https://docs.djangoproject.com/en/2.0/releases/2.0/ 

This requires us to upgrade to python 3 aswell

Issue #36, opened 2017-03-22
Upgrade to python 3


fishbacon:
Python 2.7 is reaching end of life, we should consider upgrading to python 3 if we are going to keep developing this version.

http://legacy.python.org/dev/peps/pep-0373/#id2

VirtualSatai:
Duplicate of #20

PR #142, opened 2018-02-15
Upgrade to Django 2.0 and Python 3.x

### LINKED ISSUE #20: Upgrade to Python 3
Django 1.11 (we run 1.10 now) is the last which will support Python 2, but it will as a LTS.

Django 2.0 and up is Python 3.x only, we should be prepared for that.

VirtualSatai:
Fixes #20 #36 #137


stregsystem/models.py
```diff
@@ -277,7 +276,7 @@ def calculate_alcohol_promille(self):
 
 
 class Payment(models.Model):  # id automatisk...
-    member = models.ForeignKey(Member)
+    member = models.ForeignKey(Member, on_delete=models.CASCADE)
```
VirtualSatai: The only real consideration in these commits are here, do we want to CASCADE deletions?

stregsystem/models.py
```diff
@@ -394,7 +393,7 @@ def is_active(self):
                 and not out_of_stock)
 
 class OldPrice(models.Model):  # gamle priser, skal huskes; til regnskab/statistik?
-    product = models.ForeignKey(Product, related_name='old_prices')
+    product = models.ForeignKey(Product, on_delete=models.CASCADE, related_name='old_prices')
```
VirtualSatai: As above

stregsystem/models.py
```diff
@@ -406,9 +405,9 @@ def __str__(self):
         return self.product.name + ": " + money(self.price) + " (" + str(self.changed_on) + ")"
 
 class Sale(models.Model):
-    member = models.ForeignKey(Member)
-    product = models.ForeignKey(Product)
-    room = models.ForeignKey(Room, null=True)
+    member = models.ForeignKey(Member, on_delete=models.CASCADE)
+    product = models.ForeignKey(Product, on_delete=models.CASCADE)
+    room = models.ForeignKey(Room, on_delete=models.CASCADE, null=True)
```
VirtualSatai: As above

DelusionalLogic:
We need a list over what we need to do to deploy this, since it's not a trivial deploy!

VirtualSatai:
- New venv: `virtualenv -p python3.6 venv`
- Update `/etc/apache2/sites-enabled/stregsystem.conf` to point at the new python3 virtualenv.
- Update `local.cfg` to have a new DB connection for `mysql.connector.django` since `django.db.backends.mysql` is py3 compadable
- Update apache2 wsgi to py3: `a2dismod wsgi` -> `apt install libapache2-mod-wsgi-py3` -> `a2enmod wsgi`
- Restart apache2
- Test functionality in prod

PR #144, opened 2018-03-07
Feature email payment

BoAnd:
Working code. I know, I'm awesome. This should send a email when there is made a payment.

### DIFF
--- a/stregsystem/models.py
+++ b/stregsystem/models.py
@@ -303,8 +303,8 @@ class Payment(models.Model):  # id automatisk...
             # TODO: Make atomic
             self.member.make_payment(self.amount)
             super(Payment, self).save(*args, **kwargs)
-            self.member.save()
-	          if self.member.email != "":
+            self.member.save() 
+            if self.member.email != "":
                 if '@' in parseaddr(self.member.email)[1]:
                     send_payment_mail(self.member, self.amount)            
 
--- a/stregsystem/models.py
+++ b/stregsystem/models.py
@@ -305,7 +305,7 @@ class Payment(models.Model):  # id automatisk...
             super(Payment, self).save(*args, **kwargs)
             self.member.save() 
             if self.member.email != "":
-                if '@' in parseaddr(self.member.email)[1]:
+                if '@' in parseaddr(self.member.email)[1] and self.member.want_spam:
                     send_payment_mail(self.member, self.amount)            
 
     def delete(self, *args, **kwargs):
--- a/stregsystem/models.py
+++ b/stregsystem/models.py
@@ -303,7 +303,7 @@ class Payment(models.Model):  # id automatisk...
             self.member.make_payment(self.amount)
             super(Payment, self).save(*args, **kwargs)
             self.member.save()
-            send_mail(self.member.email)            
+            send_mail(self.member, self.amount)            
 
     def delete(self, *args, **kwargs):
         if self.id:
--- a/stregsystem/models.py
+++ b/stregsystem/models.py
@@ -7,7 +7,7 @@ from django.utils import timezone
 from stregsystem.deprecated import deprecated
 from stregsystem.templatetags.stregsystem_extras import money
 from stregsystem.utils import date_to_midnight
-from stregsystem.utils import send_mail
+from stregsystem.utils import send_payment_mail
 
 def price_display(value):
     return money(value) + " kr."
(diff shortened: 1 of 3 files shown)


stregsystem/utils.py
```diff
@@ -72,3 +73,13 @@ def date_to_midnight(date):
     :return: the date as a timezone aware datetime at midnight
     """
     return timezone.make_aware(timezone.datetime(date.year, date.month, date.day, 0, 0))
+
+
+def send_mail(user_email):
+    msg = "Your money is now on your streg account"
```
VirtualSatai: Could we improve the message? 
f"Hey {member.firstname}, we've added {payment.amount} to your streg-account: '{member.username}'"

Is there an subject field? to set?

stregsystem/utils.py
```diff
@@ -72,3 +73,13 @@ def date_to_midnight(date):
...
+    try:
+       smtpObj = smtplib.SMTP('localhost')
+       smtpObj.sendmail('FIT@fklub.dk', user_email, msg)
+    except:
+       pass
```
VirtualSatai: Logging of mistakes?

stregsystem/utils.py
VirtualSatai: Maybe we could add: "If you don't want to receive more spam contact: fklub@elefsennet.dk"

BoAnd: Yes, we should improve the message. And I like your ideas.

stregsystem/utils.py
BoAnd: Yes, but where to ?

stregsystem/models.py
```diff
@@ -304,7 +304,7 @@ def save(self, *args, **kwargs):
             self.member.make_payment(self.amount)
             super(Payment, self).save(*args, **kwargs)
             self.member.save()
-            send_mail(self.member.email)            
+            send_mail(self.member, self.amount)
```
VirtualSatai: Add a condition that so that we only try to send an email if the fember has one.

stregsystem/utils.py
```diff
@@ -75,8 +75,8 @@ def date_to_midnight(date):
     return timezone.make_aware(timezone.datetime(date.year, date.month, date.day, 0, 0))
 
 
-def send_mail(user_email):
-    msg = "Your money is now on your streg account"
+def send_mail(member, payment):
+    msg = "Hey {member.firstname} we've added {payment.amount} to your streg-account: '{member.username}'"
     try:
        smtpObj = smtplib.SMTP('localhost')
        smtpObj.sendmail('FIT@fklub.dk', user_email, msg)
```
VirtualSatai: `user_email` should be `member.email` <- if that variable exists, otherwise exit?

stregsystem/utils.py
DelusionalLogic: To here: https://docs.djangoproject.com/en/2.0/topics/logging/

stregsystem/utils.py
```diff
@@ -72,3 +73,13 @@ def date_to_midnight(date):
     :return: the date as a timezone aware datetime at midnight
     """
     return timezone.make_aware(timezone.datetime(date.year, date.month, date.day, 0, 0))
+
+
+def send_mail(member, amount):
```
DelusionalLogic: This doesn't look like a generic send_mail function. Maybe it should be named something like "send_payment_mail"?

stregsystem/utils.py
```diff
@@ -72,3 +73,13 @@ def date_to_midnight(date):
     :return: the date as a timezone aware datetime at midnight
     """
     return timezone.make_aware(timezone.datetime(date.year, date.month, date.day, 0, 0))
+
+
+def send_payment_mail(member, amount):
+    msg = f"Hey {member.firstname} we've added {amount} to your streg-account: '{member.username}'"
```
VirtualSatai: Maybe prepend a subject field:
https://stackoverflow.com/questions/7232088/python-subject-not-shown-when-sending-email-using-smtplib-module?answertab=votes#tab-top

stregsystem/utils.py
```diff
@@ -76,10 +76,27 @@ def date_to_midnight(date):
...
+    msg['Subject'] = 'Streg-account payment'
+    
+    html = f"
+    <html>
+        <head></head>
+        <body>
+            <p>Hey {member.firstname}<br>
+               We've added {amount}$ to your streg-account: {member.username}<br>
+               <br>
+               If you don't want to receive more spam contact: fklub@elefsennet.dk
```
VirtualSatai: `<a href="mailto:fklub@elefsennet.dk?Subject=Klage" target="_top">fklub@elefsennet.dk</a>`

PR #154, opened 2018-04-01
Init easter fools

BoAnd:


stregsystem/utils.py
```diff
@@ -1,4 +1,5 @@
 import logging
+from datetime import date
```
hrjakobsen: This is not timezone aware. In #116  @VirtualSatai said we should use `from django.utils import timezone` and `timezone.now()`.

hrjakobsen:
It does not happen on the /sale page.

PR #158, opened 2018-04-11
The stregsystem is now future complete.

### LINKED ISSUE #157: Stregforbud should return 402 payment required
To be compliant with the HTTP status codes we need to return 402 error when fembers tries to buy a product when they do not have sufficient amount of stregdollars (stregforbud).

fishbacon:
Fixes #157


fishbacon:
Needs tests.

PR #160, opened 2018-05-29
Add MobilePay QR code screen

### LINKED ISSUE #159: Change stregforbuds page
Currently, the stregforbuds page informs the user to show up at Fridays between 12-12:30 to make payments, but since we in the TREO only accept payments by MobilePay and Cash, it should be changed to reflect that.

Here you could play around and maybe generate a QR code for MobilePay, with the amount of last payment and the username in the comment (if we ever get to write a comment again)

kimsey0:
Fixes #159

### DIFF
--- a/stregsystem/templates/stregsystem/error_stregforbud.html
+++ b/stregsystem/templates/stregsystem/error_stregforbud.html
@@ -28,7 +28,7 @@ Du kan <b>ikke</b> foretage k&oslash;b, f&oslash;r du har betalt din g&aelig;ld!
 
 <b>Din vare er ikke betalt, stil den tilbage!</b> <br />
 
-Du kan foretage indbetalinger på fredage mellem kl 12.00 og 12.30 i Jægerstuen i klynge 5 på Cassiopeia.<br />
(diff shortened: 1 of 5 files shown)


DelusionalLogic:
Could it be inline in the stregforbud error page? I think i'd like that more.

If you don't want to spend the time, this is fine too.

kimsey0:
I can add a QR code to the stregforbud error page as well? Or did you want to remove the Payment screen and the link from the user page completely? I like the idea of having those for everyone, not just those with stregforbud.

VirtualSatai:
I think having inlined and duplicated two places is a good idea. One is as the error page when you try to buy without funds (stregforbud) and the other is from ones user page like you added. But as @DelusionalLogic says, if you don't want to do more work this is good.

kimsey0:
I've factored MobilePay QR code generation out into a template tag and have added a QR code without a preset amount to the stregforbud error page. I don't think it makes sense to preset the amount so that it just covers the negative balance. Most likely, people will also want to add some funds at the same time. However, I can change this if you think otherwise.

PR #167, opened 2018-10-01
Small thing

BoAnd:


hrjakobsen:
Hacktoberfest is good for stregsystemet 😂

stregsystem/views.py
```diff
@@ -40,11 +40,11 @@ def __get_news():
 
 
 def __get_productlist(room_id):
-    l = (
-        make_active_productlist_query(Product.objects)
-                       .filter(make_room_specific_query(room_id))
+    lst = (
```
VirtualSatai: Hvorfor ikke bare return den med det samme så?

stregsystem/views.py
```diff
@@ -40,11 +40,10 @@ def __get_news():
 
 
 def __get_productlist(room_id):
-    l = (
-        make_active_productlist_query(Product.objects)
-                       .filter(make_room_specific_query(room_id))
+    return (
```
hrjakobsen: The outer parentheses seems unnecessary, but it's a very minor detail

VirtualSatai: It's only there because Jesper has a phobia of long lines.

PR #169, opened 2018-10-08
Limit number of bats

hrjakobsen:
The Pi and laptops cannot handle even 8 bats, they will be useless with 30 bats.

### DIFF
--- a/stregsystem/static/stregsystem/bat.js
+++ b/stregsystem/static/stregsystem/bat.js
@@ -29,9 +29,10 @@ spawn_bat = function () {
 };
 
 d = new Date();
+var NUM_BATS = 2;
 
 if(d.getMonth() === 9){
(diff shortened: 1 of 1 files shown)


VirtualSatai:
I dislike this change

hrjakobsen:
Then you should donate the raspberry pi this time 😂

DelusionalLogic:
Please revert this. My workflow depended in the number of bats. Now that there's too few bats, chrome doesn't eat as much memory. With chrome not eating memory, the OOM Reaper doesn't kill the processes i need killed periodically.

I guess you don't care about your downstream users :kaaba:

hrjakobsen:
I think we should we should use `log2(day)` instead of limiting it to two. Then it would still scale with days but hopefully not kill the Pi.

PR #170, opened 2018-10-08
Move html message from code to template file instead.

BoAnd:


hrjakobsen:
Not relevant due to #185 being merged, but can be recreated for the other implementation.

PR #172, opened 2018-10-24
Optimise bats

Zinoex:


stregsystem/static/stregsystem/bat.js
```diff
@@ -1,38 +1,49 @@
...
+        b = y;
         setTimeout(A,d);
     }
-    setTimeout(A, r() * 3e3);
+    setTimeout(A, Math.random() * 3e3);
 };
 
 d = new Date();
-var NUM_BATS = 2;
+var NUM_BATS = 200;
```
hrjakobsen: We probably shouldn't show 200 bats at once.

Zinoex: Probably right xD
It is just test code, I forgot to remove.

DelusionalLogic: > We probably shouldn't show 200 bats at once.

At least 400 bats are needed

```suggestion
var NUM_BATS = 400;
```

VirtualSatai:
@hrjakobsen any reason for not getting this in ASAP and deploy before we crash a pi? Spooktober is upon us!

hrjakobsen:
There's already 3 bats on the pi even though I thought there would only be 2. I'll merge it tomorrow (maybe)

stregsystem/static/stregsystem/bat.js
```diff
@@ -1,38 +1,48 @@
...
+    var z = document.createElement('div');
+    var zs = z.style;
+    var a = window.innerWidth * Math.random();
+    var b = window.innerHeight * Math.random();
+    z.classList.add("bat");
+    z.appendChild(v);
+    v.appendChild(s);
+    v.width = "48";
+    v.height = "48";
+    v.autoplay = true;
```
hrjakobsen: We should also add `v.muted = true;` since chrome will otherwise not autoplay

hrjakobsen: I did that in 4a9676ddeeb6efe85a35207bcad5958a9161309b

Issue #163, opened 2018-09-30
Top 10 sales view page 


BoAnd:
Somebody would like to have a top 10 high score of sales "https://github.com/f-klubben/stregsystemet/blob/next/stregreport/views.py#L217" integrate with Kiosk app.

VirtualSatai:
This has previously been discussed, and I think the conclusion was that something like this should be opt-in. But I'm sure you can find an old discussion somewhere.

hrjakobsen:
I'm not sure about the implementation of this. If websites in the kiosk were a thing, then it would be simple. But with only images, I don't know.. Generating images and keeping track of a kiosk item seems overly complicated.

PR #177, opened 2019-05-23
Add simple user statistics about sales

kimsey0:
As shown by #163, there is some interest in statistics, but concern about making global high-score lists. This pull request adds simple sale statistics (count, sum, average, min, and max) to the user info page. That way, users can themselves decide if and when they look and compare.

If this feature is successful, more statistics can be added, such as most bought item (with count), average time between purchases, earliest purchase in the morning or latest at night, etc.

The feature has been tested for users with zero purchases, where it just shows zeroes for all statistics.

### DIFF
--- a/stregsystem/templates/stregsystem/menu_userinfo.html
+++ b/stregsystem/templates/stregsystem/menu_userinfo.html
@@ -86,5 +86,35 @@ Ingen indbetalinger!
    </tr>
 </table>
 
+<br />
+
+<center><b>Vi har beregnet følgende statistikker om dig</b></center>
+<table border=1 width="100%">
+   <tr>
+      <th align="left">Felt</th>
+      <th align="right">Værdi</th>
+   </tr>
+   <tr>
+      <td>Totale antal køb</td>
+      <td align="right">{{sale_statistics.number_sales}}</td>
+   </tr>
+   <tr>
+      <td>Total købspris</td>
+      <td align="right">{{sale_statistics.total_price|money}}</td>
+   </tr>
+   <tr>
+      <td>Gennemsnitlig købspris</td>
+      <td align="right">{{sale_statistics.average_price|money}}</td>
+   </tr>
+   <tr>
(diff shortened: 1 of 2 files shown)


hrjakobsen:
I don't think we should show this information. I think knowing that you spent 6000,- on stregsystemet will encourage you to stop buying stuff, even though it's been spent over multiple years. Minimum purchase will always be 0 for a free beer (when this is tracked). Maximum purchase will always be Fyttetur. I don't see the use for average purchase at all.

kimsey0:
I'll be happy to remove minimum, maximum and average if you don't find them relevant.

I don't see the problem in showing total purchases. If seeing that you spent 6000 on Stregsystemet makes you think that you should stop buying stuff, maybe you actually should stop buying stuff. Stregsystemet is not for profit, but for the benefit of F-klubben members. Products are sold at (or as close to) cost as possible. Losing, say, 10 % of the revenue shouldn't damage the viability. And I don't think that's going to happen. I know my number will be large, but I want to see it, and I know it won't have any influence on my further purchases.

hrjakobsen:
How about showing number of purchases instead? Or maybe even number of purchases per category? I think I'm not getting what the point of this feature is. I see the point of having highscores as mentioned in #163, as it will be fun and people could compete if they so desire. Showing how much money you spent during 4-5 years seems irrelevant to me. 

If other people disagree with me, we can for sure accept the pull request, but I will wait and see what other people says first.

kimsey0:
I feel differently about total purchase price. To me, that's the most interesting number and the one that made me spend 30 minutes creating this pull request. 😉
I can't speak for everyone, but I know my group mates want to see that number as well. If you don't want to see it and don't think you could overlook it when on the info page, we could even hide it beneath a Reddit-style spoiler tag.

The code in the pull request already shows number of purchases. I'm happy to implement number of purchases per category or other fun metrics as well, either in this pull request or in another one after this one is merged.

PR #178, opened 2019-09-10
Avoid duplicate member usernames

hrjakobsen:
This PR adds error messages when new fembers are created with usernames that already exists. If an existing member is updated with a duplicate username, a warning message is shown for backwards compatibility with duplicate usernames. 

This feature was requested by TREO'en.

### DIFF
--- a/stregsystem/admin.py
+++ b/stregsystem/admin.py
@@ -1,5 +1,7 @@
 from django.contrib import admin
+from django import forms
 from django.contrib.admin.views.autocomplete import AutocompleteJsonView
+from django.contrib import messages
 
 from stregsystem.models import (
     Category,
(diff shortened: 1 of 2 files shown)


stregsystem/tests.py
```diff
@@ -31,6 +35,13 @@
     price_display
 )
 
+class MockRequest():
```
jenrik: Where is MockRequest and MockFormChange used? I can't spot any references to them

hrjakobsen: Good catch, they were used in some code that I ended up replacing in the Member Admin tests. I'll remove that.

stregsystem/admin.py
```diff
@@ -182,6 +198,12 @@ class MemberAdmin(admin.ModelAdmin):
         })
     )
 
+    def save_model(self, request, obj, form, change):
+        if 'username' in form.changed_data and change:
```
jenrik: Why are the username checked here if it's already being validated in the form class?

hrjakobsen: Because the form cannot send messages, since it has no access to request, so it can only invalidate the form, and cancel the creation/update. So only in the save_model can we access the request and send a warning message.

This is for when a member is updated, and the username is duplicate. If we don't allow this, old data cannot be updated without errors.

PR #179, opened 2019-09-22
Reformat, cleanup, tox- and isort-fix

cogitantium:
In an attempt to get to know the stregsystem-codebase, I've done the following:
- Entire project (excluding migrations and non-py files) has been reformatted to adhere to ~~flake8~~ black and PEPs except E501. I realise that these changes are less readable in some cases. ~~I tried to strike a balance between readability and adherence to flake8.~~
- Did minor cleanup, such as unused variables, redundant parentheses, wrong function signatures, etc.
~~- Found an [issue with isort-check](https://github.com/timothycrosley/isort/issues/704), where third party libraries are not sorted correctly. Fixed by adding `requirements.txt` to isort's dependencies. Now isort works when run through tox. Applied suggested changes to codebase.~~
~~- Bumped python from 3.6 to 3.7 in tox's envlist. Required to be able to run on my machine. Let me know if wrong.~~


stregsystem/models.py
```diff
@@ -181,7 +179,9 @@ def __unicode__(self):
         return self.__str__()
 
     def __str__(self):
-        return active_str(self.active) + " " + self.username + ": " + self.firstname + " " + self.lastname + " | " + self.email + " (" + money(self.balance) + ")"
+        return active_str(
+            self.active) + " " + self.username + ": " + self.firstname + " " + self.lastname + " | " + self.email \
+            + " (" + money(self.balance) + ")"
```
BoAnd: I think we should use f-strings instead here. (https://www.python.org/dev/peps/pep-0498/).

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py36, flake8, isort-check
+envlist = py37, flake8, isort-check
```
BoAnd: When we bump the version. We should also update the README.md (https://github.com/f-klubben/stregsystemet/blob/next/README.md)

hrjakobsen: Also we need to make sure this is consistent with the server python version

kiosk/views.py
```diff
@@ -39,20 +39,11 @@ def find_next_image_real(request, item_id):
     # Get the item at the index, trust that Django does this smartly.
     try:
         next_item = (
-            KioskItem.objects
```
hrjakobsen: I think this chaining actually looks better than what flake8 seems to think. It's much more readable i think.

tox.ini
cogitantium: @BoAnd resolved README.md versioning in cfbc192? I suppose it's only Anaconda which specifies minor version of Python.

@hrjakobsen do you mean the Python version which is running in production? I can't really find any other references in the codebase.

hrjakobsen: Server version on the production server is 3.6.

cogitantium: Okay, so what do? Upgrading Python in production or keeping 3.6?

stregsystem/models.py
cogitantium: Resolved in bea4cb9

cogitantium:
`flake8` is pretty weird when breaking chaining over multiple lines, so I looked for alternatives and found [black](https://github.com/psf/black). I like it so far - this also allows us to get rid of isort-check as black handles import too.

E.g. see previous `kiosk/views.py`, as discussed with @hrjakobsen:
```python
    try:
        next_item = (
            KioskItem.objects
            .filter(active=True)
            .order_by('ordering', 'id')
            .filter(
                Q(ordering__gt=item.ordering)
                | (Q(ordering=item.ordering) & Q(id__gt=item.id))
            )[0]
        )
```

Becomes with black:
```python
    try:
        next_item = (
            KioskItem.objects.filter(active=True)
            .order_by("ordering", "id")
            .filter(Q(ordering__gt=item.ordering) | (Q(ordering=item.ordering) & Q(id__gt=item.id)))[0]
        )
```
Which I think is a livable compromise.

I guess this PR turned into a request to replace the tooling and apply the changes suggested by these. Note that I updated `CONTRIBUTING.md` to reflect this change.

Note also, that the `tox.ini` specifies py37 now. It is undecided whether this is what we're going to do.

kiosk/views.py
cogitantium: Agreed. See my latest comment on a possible fix.

tox.ini
jenrik: I don't think we're gonna be able to upgrade python beyond 3.6 until Ubuntu decides to ship a newer version and that is not gonna be until the next long term service release I think

hrjakobsen:
I'm not sure if i'm misunderstanding something, but when i run `black` (which should be added to `requirements.txt` by the way) on the code from this commit, it still changes the code and in fact  it changes the above example you provided to:
```diff
--- kiosk/views.py	2019-10-08 10:45:04.782957 +0000
+++ kiosk/views.py	2019-10-08 10:50:57.763010 +0000
@@ -33,11 +33,14 @@
     # Get the item at the index, trust that Django does this smartly.
     try:
         next_item = (
             KioskItem.objects.filter(active=True)
             .order_by("ordering", "id")
-            .filter(Q(ordering__gt=item.ordering) | (Q(ordering=item.ordering) & Q(id__gt=item.id)))[0]
+            .filter(
+                Q(ordering__gt=item.ordering)
+                | (Q(ordering=item.ordering) & Q(id__gt=item.id))
+            )[0]
         )
     except IndexError:
         next_item = KioskItem.objects.filter(active=True).order_by("ordering", "id")[0]
     response_data = {"id": next_item.id, "url": next_item.image.url}
     return HttpResponse(json.dumps(response_data), content_type="application/json")
```

I thought this PR had run `black` on all files? And if it has, is the `black` standard changing too rapidly to keep up with?

cogitantium:
@hrjakobsen as discussed on Slack, you seem to have forgot the `-l 120` option, which is pretty much the only option `black` takes.
When I ran `black -t py36 -l 120 --diff --exclude "migrations" stregsystem/ stregreport/ kiosk/` I got no changes on b546b34c6eaafce9bc5cd7f23c5a02b8a0214404.

500c491b70d35b3dc521b224d4a0065cdeae0383 adds `black` 19.3b0 to `requirements.txt`. I don't think we should specify this-or-greater version, unless we're happy with the changes that `black`s devs makes.

VirtualSatai:
Is this obsolete given #229 ?

cogitantium:
> Is this obsolete given #229 ?

Yes, precisely

PR #182, opened 2019-10-01
Hide username after 5 seconds

joandrsn:
Hides the username after 5 seconds. Stregsystemet is supposed to be trust based, but you can see usernames after purchase. This also might not be GDPR compliant. As such here is a feature to hide the username after 5 seconds.

It simply replaces the username with "x"'s. 

If it were to reload the site, while another user was inputting their order, it would be removed.


joandrsn:
> 
> 
> The action is still visible after we mask the username. I think it would be better if we removed the entire action line after a short interval

The reason for this is to not move the price list. It could be a potential annoyance for the next user if he/she is reading the list and it moves.

hrjakobsen:
I think anonymizing it is fine. Otherwise you might miss the notification and wonder if you've purchased anything. This way you will know for sure, even if the username is removed.

stregsystem/static/stregsystem/gdpr.js
```diff
@@ -0,0 +1,16 @@
...
+*/
+
+var username_element = document.getElementById("username");
+if (username_element !== null) {
+
+  setTimeout(function () {
+
+    var username_element = document.getElementById("username");
+    if (username_element !== null) {
+      username_element.innerText = "x".repeat(username_element.innerText.length);
```
cogitantium:
```suggestion
      username_element.innerText = "xxxxxxxx";
```

If we're anonymising, then we might as well do it properly. I suggest statically printing `xxxxxxxx` instead.

_Also, I have an idea for a meme on this. I'll PR that later_

stregsystem/static/stregsystem/gdpr.js
```diff
@@ -0,0 +1,16 @@
+/*
+Hides the username after 5 seconds of showing it.�
```
hrjakobsen: What is the last character here? I just see a <?>.

PR #184, opened 2019-10-02
Make the kiosk admin page more Danish

ScumbagDog:
Currently the kiosk admin page has two options for making marked items active and inactive, but these are both in English! Clearly we should make them Danish instead, and this PR fixes that.


VirtualSatai:
"items" isn't really a danish word, otherwise this looks good to me :D

PR #152, opened 2018-03-20
Feature sign up mail

BoAnd:
Still missing mail body. But please review.

PR #185, opened 2019-10-04
Added welcome mail

joandrsn:
This is adding a welcome mail when saving a new member.

The solution utilises [Django Signals](https://docs.djangoproject.com/en/2.2/topics/signals/) to catch the event where a new member is saved and then sends an email.

If the solution is acceptable, `email.py` could have all the email communication. 

Debugging SMTP server:
Using this command:
```
sudo python -m smtpd -n -c DebuggingServer localhost:25
```
Provided from [Serverfault](https://serverfault.com/questions/207619/how-to-setup-a-fake-smtp-server-to-catch-all-mails), you can see that a mail is received.

Any feedback is welcome.

### DIFF
--- a/stregsystem/__init__.py
+++ b/stregsystem/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = 'stregsystem.apps.StregConfig'
--- a/stregsystem/apps.py
+++ b/stregsystem/apps.py
@@ -1,5 +1,11 @@
 from django.apps import AppConfig
 
+from django.db.models.signals import post_save
 
 class StregConfig(AppConfig):
     name = 'stregsystem'
+
+    def ready(self):
+        from .models import Member
+        from stregsystem.signals import after_member_save
+        post_save.connect(after_member_save, sender=Member)
--- a/stregsystem/apps.py
+++ b/stregsystem/apps.py
@@ -1,11 +1,11 @@
 from django.apps import AppConfig
 
 from django.db.models.signals import post_save
+from stregsystem.signals import after_member_save
 
 class StregConfig(AppConfig):
     name = 'stregsystem'
 
     def ready(self):
-        from .models import Member
-        from stregsystem.signals import after_member_save
+        from stregsystem.models import Member
         post_save.connect(after_member_save, sender=Member)
--- a/stregsystem/mail.py
+++ b/stregsystem/mail.py
@@ -4,6 +4,7 @@ import logging
 
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
+from .utils import money
 
 
 logger = logging.getLogger(__name__)
@@ -18,21 +19,27 @@ def send_email(mailadress, msg_string):
 
 def send_welcome_mail(member):
     msg = MIMEMultipart()
+    
+    if member.balance != 0:
+        balancemsg = f"""Din stregkonto har en balance på {money(member.balance)}"""
+    else:
+        balancemsg = f"""Bemærk at du skal indsætte penge på din stregkonto før at du kan købe noget på strandvejen.<br>
+        Dette kan gøres ved at sende en yderligere betaling på MobilePay til Fklubben, med det beløb som du gerne vil have indsat."""
+    
     html = f"""
     <html>
         <head></head>
         <body>
-            Hej {member.firstname}!
+            Hej {member.firstname}!<br><br>
             Velkommen som fember i Fklubben.
-            Du har nu en brugerkonto oprettet. Dit brugernavn er {member.username}. 
+            Du har nu en brugerkonto oprettet. Dit brugernavn er {member.username}.<br><br>
 
-            De 200 kr du har givet for at blive medlem betaler for din adgang til fredagsfranskbrød, hver onsdag.
-            Når du møder op til fredagsfranskbrød, så kan du opleve en razzia. Her skriver du dit brugernavn ind ({member.username}), for at sikre at du er medlem af fklubben.
+            De 200 kr du har givet for at blive medlem betaler for din adgang til fredagsfranskbrød, hver onsdag.<br><br>
 
-            Med dit brugernavn kan du også købe drikkevarer fra strandvejen.
-            Bemærk at du skal indsætte penge på din stregkonto før at du kan købe noget på strandvejen.
-            Dette kan gøres ved at sende en yderligere betaling på MobilePay til Fklubben, med det beløb som du gerne vil have indsat.
-            Priserne kan ses i stegsystemet på strandvejen. 
+            Med dit brugernavn kan du også købe drikkevarer fra Strandvejen. 
+            Strandvejen er det lille køkken i klynge 2, stueetagen, nær Jægerstuen med de fire køleskabe og stregsystemets terminal.<br>
+            {balancemsg}<br>
+            Priserne kan ses i stegsystemet på strandvejen.<br><br><br>
 
 
             Mvh,<br>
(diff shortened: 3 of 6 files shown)


stregsystem/mail.py
```diff
@@ -0,0 +1,47 @@
...
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+
+
+logger = logging.getLogger(__name__)
+
+def send_email(mailadress, msg_string):
+    try:
+        smtpObj = smtplib.SMTP('localhost', 25)
+        smtpObj.sendmail('treo@fklub.dk', mailadress, msg_string)
```
jenrik: Should the default email really always be Treo? I think it would be better if we at least made it an argument with a default value. Preferably I would like it to be a config option

stregsystem/mail.py
```diff
@@ -0,0 +1,47 @@
...
+
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+
+
+logger = logging.getLogger(__name__)
+
+def send_email(mailadress, msg_string):
+    try:
+        smtpObj = smtplib.SMTP('localhost', 25)
```
jenrik: This should be a option such that we can configuring it differently in a dev environment

stregsystem/__init__.py
```diff
@@ -0,0 +1 @@
+default_app_config = 'stregsystem.apps.StregConfig'
```
jenrik: Why is this relevant for this pull request?

stregsystem/apps.py
```diff
@@ -1,5 +1,11 @@
 from django.apps import AppConfig
 
+from django.db.models.signals import post_save
 
 class StregConfig(AppConfig):
     name = 'stregsystem'
+
+    def ready(self):
+        from .models import Member
+        from stregsystem.signals import after_member_save
```
jenrik: Why is this import statement not at the top of the file?

stregsystem/signals.py
```diff
@@ -0,0 +1,14 @@
+
+from django.db.models.signals import post_save
+from django.db.models import F
+from django.dispatch import receiver
+
+def after_member_save(sender, instance, created, **kwargs):
+    
+    from .mail import send_welcome_mail
```
jenrik: Why isn't this import statement not at the top of the file?

jenrik:
Crosslinking to #152 since it's related

stregsystem/mail.py
```diff
@@ -0,0 +1,47 @@
...
+        <body>
+            Hej {member.firstname}!
+            Velkommen som fember i Fklubben.
+            Du har nu en brugerkonto oprettet. Dit brugernavn er {member.username}. 
+
+            De 200 kr du har givet for at blive medlem betaler for din adgang til fredagsfranskbrød, hver onsdag.
+            Når du møder op til fredagsfranskbrød, så kan du opleve en razzia. Her skriver du dit brugernavn ind ({member.username}), for at sikre at du er medlem af fklubben.
+
+            Med dit brugernavn kan du også købe drikkevarer fra strandvejen.
+            Bemærk at du skal indsætte penge på din stregkonto før at du kan købe noget på strandvejen.
```
cogitantium: This welcome mail should print the current balance of the member. However, if this mail is send immediately after user-creation, then the Treo-admin may not have inserted their balance yet.

In that case a mail should follow shortly with the given balance-topoff and this welcome-text should reflect that.

stregsystem/mail.py
```diff
@@ -0,0 +1,47 @@
...
+            Dette kan gøres ved at sende en yderligere betaling på MobilePay til Fklubben, med det beløb som du gerne vil have indsat.
+            Priserne kan ses i stegsystemet på strandvejen.
```
cogitantium:
```suggestion
            Priserne kan ses i stegsystemet på Strandvejen. 
```

stregsystem/__init__.py
joandrsn: I've included it, since if not, the `ready` function will not run. More reading: https://docs.djangoproject.com/en/2.2/ref/applications/#django.apps.AppConfig.ready

stregsystem/mail.py
joandrsn: Agree, but I simply used the same configuration as was already there. I agree, this should be a setting, but maybe someone else wants to do this? Or I could include it now, if you want?

stregsystem/apps.py
joandrsn: Because I cannot import them in the module-level. I can only import them in the `ready`-function. More reading: https://docs.djangoproject.com/en/2.2/ref/applications/#django.apps.AppConfig.ready

stregsystem/signals.py
joandrsn: Once again, because I cannot import them at the module-level.

stregsystem/mail.py
joandrsn: See answer below

stregsystem/mail.py
```diff
@@ -0,0 +1,54 @@
...
+        <body>
+            Hej {member.firstname}!<br><br>
+            Velkommen som fember i Fklubben.
+            Du har nu en brugerkonto oprettet. Dit brugernavn er {member.username}.<br><br>
+
+            De 200 kr du har givet for at blive medlem betaler for din adgang til fredagsfranskbrød, hver onsdag.<br><br>
+
+            Med dit brugernavn kan du også købe drikkevarer fra Strandvejen. 
+            Strandvejen er det lille køkken i klynge 2, stueetagen, nær Jægerstuen med de fire køleskabe og stregsystemets terminal.<br>
+            {balancemsg}<br>
```
cogitantium:
```suggestion
            Din stregkonto-saldo: {balancemsg}<br>
```

cogitantium: Resolved in c281c7b

stregsystem/mail.py
jenrik: If we're already using hardcoded addresses for the mail server, then it's fine as long as we create an issue such that it can be addressed later

stregsystem/mail.py
jenrik: Which answer are you referring to? The preexisting use of hardcoded values?

joandrsn: Correct, just below :)

stregsystem/mail.py
joandrsn: Done #187

joandrsn:
I've disabled the mails using a `DiscoverRunner` (Read here: [https://docs.djangoproject.com/en/2.2/topics/testing/advanced/](DiscoverRunner)).
It utilises the Django settings to make a variable called `TEST_MODE`. When running the test suite, this is defined and thus exits the functions sending mails.

I would actually really like to use Django's built in [mail sender](https://docs.djangoproject.com/en/2.2/topics/email/), as this already has support for defining a dummy backend. Maybe this could be corrected in another revision?

stregsystem/mail.py
```diff
@@ -0,0 +1,47 @@
...
+    try:
+        smtpObj = smtplib.SMTP('localhost', 25)
+        smtpObj.sendmail('treo@fklub.dk', mailadress, msg_string)
+    except Exception as e:
+        logger.error(str(e))
+
+
+def send_welcome_mail(member):
+    msg = MIMEMultipart()    
+    html = f"""
```
VirtualSatai: Should we offer an english version or go danish only?

jenrik: I think danish is good enough for now, but appending an english version would be a nice future addition

cogitantium: Agreed. This is also how most of the public information from F-klubben is posted.

Rene4100:
This seems to be similar to #152  is that right

stregsystem/mail.py
hrjakobsen: I've appended an english version to the email in c79a4f2

hrjakobsen: image

PR #186, opened 2019-10-04
Close P tag in payment email

joandrsn:
Closes the P tag in the payment mail


hrjakobsen:
While I agree that we should close the tag, isn't omitting the closing tag of a P element allowed in HTML5?

PR #190, opened 2019-10-29
Add Foobar razzia

hrjakobsen:
This adds a new type of razzia, where it is allowed to register multiple times. Most migration files are for data persistence reason (had to migrate the data manually since we are changing a many-to-many field to a many-to-many with a `through`). 

This could probably have been made simpler, but I think this is a more extensible approach, if other razzias need to be added later.


hrjakobsen:
Oh, and I forgot to mention that this was done upon request from Forstyrrelsen/Foobar.

stregreport/views.py
```diff
@@ -39,49 +39,61 @@ def sales(request):
 sales = staff_member_required(sales)
 
 
-def bread(request, razzia_id):
+def razzia(request, razzia_id, razzia_type=BreadRazzia.BREAD, title=None):
     if request.method == 'POST':
-        return bread_view(request, razzia_id, request.POST['username'])
+        return razzia_view_single(request, razzia_id, request.POST['username'], razzia_type=razzia_type, title=title)
```
jenrik: wouldn't it be better if we used a Django form instead of accessing `request.POST`?

stregreport/views.py
```diff
@@ -39,49 +39,61 @@ def sales(request):
...
-            if not already_used:
-                razzia.members.add(member)
-                razzia.save()
-
-    return render(request, 'admin/stregsystem/razzia/bread.html', locals())
+            entries = list(razzia.razziaentry_set.filter(member__pk=member.pk).order_by('-time'))
+            already_used = len(entries) > 0
+            if already_used:
+                entry = entries[0]
+            if not already_used or razzia_type == BreadRazzia.FOOBAR:
```
jenrik: I don't like that we don't hardcoding a check for `BreadRazzia.FOOBAR`, I would rather that we have a way to check if you're allowed to check in multiple times for a certain type of razzia

stregsystem/migrations/0009_auto_20191028_2057.py
```diff
@@ -0,0 +1,23 @@
...
+            model_name='member',
+            name='gender',
+            field=models.CharField(choices=[('U', 'Unknown'), ('M', 'Male'), ('F', 'Female')], max_length=1),
+        ),
+        migrations.AlterField(
+            model_name='member',
+            name='year',
+            field=models.CharField(default='2019', max_length=4),
+        ),
+    ]
```
jenrik: This seem unrelated to this PR, why has it been included?

hrjakobsen: There is a dependency of migrations from stregreport to stregsystem, so I couldn't get Django to generate migrations for stregreport without also creating for some changes in stregsystem that was never migrated. This should probably be a commit/PR of its own, but if this PR is merged anyways, I don't think it's too much of a problem.

hrjakobsen:
I'm merging this now so it can be used for the event tomorrow.

PR #196, opened 2020-02-20
Feature easter egg username replacements

cogitantium:
Very similar to #155 but utilises the username-hider feature #182, so the replaced username is on-screen until the next user approaches. Therefore, more people should be able to see the replaced usernames.

Allowing the feature to trigger at another interval could make it more than a once-a-year occurrence. Each Friday, maybe? 

The ordinary replacement 'xxxxxxxx' is replaced with '[username_hidden]' for clarity _but mostly personal preference_.

### DIFF
--- a/stregsystem/static/stregsystem/gdpr.js
+++ b/stregsystem/static/stregsystem/gdpr.js
@@ -6,10 +6,17 @@ Simply replaces the letters with 'x'.
 var username_element = document.querySelectorAll(".username");
 
 if (username_element.length > 0) {
-  setTimeout(function () {
-    var username_element = document.querySelectorAll(".username");
-    for (var i = 0; i < username_element.length; i++) {
-    	username_element[i].innerText = "xxxxxxxx";
-    }
-  }, 5000);
+    setTimeout(function () {
+        var username_element = document.querySelectorAll(".username");
+        var replaced_username = "[username_hidden]";
+        var date = new Date();
+        // Only do username memes on April Fools
+        if (date.getDate() === 1 && date.getMonth() + 1 === 4) {
+            const easter_egg_names = ["user", "admin", "root", "alan_turing", "hackerman", "hans_huttel"];
+            replaced_username = easter_egg_names[Math.floor(Math.random() * easter_egg_names.length)];
+        }
+        for (var i = 0; i < username_element.length; i++) {
+            username_element[i].innerText = replaced_username;
+        }
+    }, 5000);
 }


stregsystem/static/stregsystem/gdpr.js
```diff
@@ -6,10 +6,17 @@ Simply replaces the letters with 'x'.
...
-    	username_element[i].innerText = "xxxxxxxx";
-    }
-  }, 5000);
+    setTimeout(function () {
+        var username_element = document.querySelectorAll(".username");
+        var replaced_username = "[username_hidden]";
+        var date = new Date();
+        // Only do username memes on April Fools
+        if (date.getDate() === 1 && date.getMonth() + 1 === 4) {
+            const easter_egg_names = ["user", "admin", "root", "alan_turing", "hackerman", "hans_huttel"];
```
hrjakobsen: Remove `hans_huttel` and keep only the 'generic' names

cogitantium: fedc04f fixes this

stregsystem/static/stregsystem/gdpr.js
```diff
@@ -11,8 +11,8 @@ if (username_element.length > 0) {
         var replaced_username = "[username_hidden]";
         var date = new Date();
         // Only do username memes on April Fools
-        if (date.getDate() === 1 && date.getMonth() + 1 === 4) {
-            const easter_egg_names = ["user", "admin", "root", "alan_turing", "hackerman", "hans_huttel"];
+        if (date.getDate() === 21 && date.getMonth() + 1 === 2) {
```
hrjakobsen: On my calendar, that is not the first of April :)

cogitantium: :upside_down_face: fixed in b41abee

cogitantium:
I suggest doing these replacements on every Friday as it otherwise likely won't ever be seen.

stregsystem/static/stregsystem/gdpr.js
```diff
@@ -12,7 +12,10 @@ if (username_element.length > 0) {
         var date = new Date();
         // Only do username replacements on Fridays
         if (date.getDay() === 5) {
-            const easter_egg_names = ["user", "admin", "root", "alan_turing", "hackerman", "foo", "bar"];
+            const easter_egg_names = ["user", "admin", "root", "alan_turing", "dijkstra", "knuth", "hackerman", "foo",
+                "bar", "Robert\'); DROP TABLE fembers;--", "hunter2", "correcthorsebatterystaple",
```
hrjakobsen: Little Bobby Tables ❤️

cogitantium:
Turns out that [Math.random returns a range inclusive of 0 but exclusive of 1](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random). This is fixed in 6c9171d.

stregsystem/static/stregsystem/gdpr.js
```diff
@@ -6,14 +6,24 @@ Simply replaces the letters with 'x'.
...
-    for (var i = 0; i < remainingbalance_element.length; i++) {
-      remainingbalance_element[i].innerText = "xx.xx";
-    }
-  }, 5000);
+    setTimeout(function () {
+        var username_element = document.querySelectorAll(".username");
+        var replaced_username = "[brugernavn skjult]";
+        var date = new Date();
+        // Only do username replacements on Fridays
+        if (date.getDay() === 5) {
```
krestenlaust, line 14:
```suggestion
        if (date.getMonth() === 3 && date.getDay() == 1) {
```

I think having it every friday will kill the joke, and would create a lot of confusion. Having it on april fools would make people confused but they'd be able to reason with it. Maybe extend the easter egg to the first monday after april fools to make sure it will go in effect

krestenlaust (changes requested):

stregsystem/static/stregsystem/gdpr.js
falkecarlsen: Having it once a year is too short of a timeframe for it to be noticed, as most fembers doesn't stay around to look at the replaced 'xxxxxxxx'.

The first Friday of every month (for example) gives ~4 days per semester for fembers to notice it.

JakobTopholt: I agree with Falke

PR #197, opened 2020-03-05
Require permission to perform breadrazzia, access sales reports and perform batch payments

hrjakobsen:


treo/settings.py
```diff
@@ -201,4 +201,7 @@
     '//cdnjs.cloudflare.com/ajax/libs/select2/4.0.4/css/select2.min.css'
 )
 
-TEST_RUNNER = 'stregsystem.utils.stregsystemTestRunner'
\ No newline at end of file
+TEST_RUNNER = 'stregsystem.utils.stregsystemTestRunner'
+
+LOGIN_REDIRECT_URL = '/admin/login'
+LOGIN_URL = '/admin/login'
```
hrjakobsen: These changes are made since `@permission_required` redirects to the login page if user is not logged in, and the default loginpage is /accounts/login instead of /admin/login that we are using.

stregsystem/templates/admin/index.html
```diff
@@ -66,13 +70,23 @@
                 <td>&nbsp;</td>
                 <td>&nbsp;</td>
             </tr>
+        </table>
+    </div>
+
+    {% endif %}
```
jenrik:
```suggestion
  {% endif %}
```

PR #199, opened 2020-03-06
Update freezegun dependency

hrjakobsen:
Update freezegun to use version instead of git commit and work with python 3.8


hrjakobsen:
I've updated this PR as I still think its relevant to have it work with python 3.8 and to not use a git commit as version in requirements.txt.

Issue #203, opened 2020-09-08
Integration to MobilePay Transaction Reporting API


joandrsn:
It would be awesome to integrate to the [MobilePay Transaction Reporting API](https://developer.mobilepay.dk/transactionreporting-api).

This could potentially be used for scenarios as these:
* Ping the treo on mail/slack when a new payment is ready (Maybe only twice a day to reduce stress in the treo)
* Automatically add funds to members balance

In order to be sure that each payment have been seen - the existing payments could have a transaction ID on it so that we know which payments in MobilePay that have been seen.

cogitantium:
Yes, very good idea!

I've spoken with Forstyrrelsen about this a while ago and we must be very sure that it functions correctly if any payments are automised at all.
A django-page that tries to do this join from Mobilepay transactions CSV or the API could be a good aid for the Treo. Features such as marking a Mobilepay transaction as 'completed' and 'pending' would help in the workflow of processing payments for the Treo.

cogitantium:
An heuristic for automatically matching payments to fembers could be to use telephone numbers to users. They're not available initially but an existing match indicates that this user has this phone number.

Only the last four digits are available from mobilepay but I believe those are random and _maybe_ uniformly distributed.

hrjakobsen:
Don't people already write their username in the message? Matching the phone number would not allow paying for others.

cogitantium:
Yes, absolutely. The best values to match for is the comment. I'm thinking exact match, then `longest common substring .len/ comment.len`, and lastly just fuzzy match confidence. 
If it's an exact match, we're pretty sure - even more so if name of payment matches fember user found.

hrjakobsen:
Only exact match should be used for automatic matching. Otherwise it should be a manual task.

cogitantium:
I should've elaborated. Matching can be done automatic but I agree that automatically submitting payments on less than exact matches is unwise.
The plan is to use the resulting confidence to suggest a user to the Treo-member handling payments, which they can then edit or sign off on. Very simplified, the UI could turn out something like the following, where the member-column acts as the existing payment-widget for fuzzily searching for a fember but in this case pre-filled with suggestion. Exact matches are pre-checked in the approval-column. 
```
Amount | Comment            | Member | Confidence | Approve
42       tester               tester   100%         [x]
42       tester (er indsat)   tester   35%          []

[Submit payments]
```
_I'm no IxD'er, suggestions welcome_

[...]

PR #204, opened 2020-09-13
Feature MobilePay transaction integration

cogitantium:
Processing payments as a Treo can be frustrating as you're referencing two lists; payments on stregsystemet and transactions from MobilePay. Sometimes Treos do payments in reverse order, making it harder to figure out which transactions are processed and which are missing. Sometimes multiple Treos are making them concurrently and I unfortunately haven't found a _human lock_ library for Python. 

Maybe this can be fully automated for 'correct' MobilePay-comments in the future when the [MobilePay Transactions API ](https://developer.mobilepay.dk/node/3832)has been set up. This has also been discussed in #203.
Meanwhile, MobilePay offers functionality to download a CSV containing the same information. This PR adds functionality to upload, parse, and create `MobilePayment` objects automatically for these transactions along with a utility for processing these transactions, creating ordinary `Payments` when approved by a Treo.

Thanks @joandrsn for figuring out MobilePay API import :ok_hand: 

image


hrjakobsen:
Expected human lock library. 10/10 disappoint.

stregsystem/admin.py
```diff
@@ -244,10 +250,41 @@ def get_amount_display(self, obj):
     get_amount_display.admin_order_field = "amount"
 
 
+class MobilePaymentAdmin(admin.ModelAdmin):
+    list_display = ('get_username', 'payment', 'customer_name', 'comment', 'member_guess', 'timestamp', 'transaction_id', 'get_amount_display', 'approval')
```
BoAnd, line 254: Maybe we should split this line. Like 
```
list_display = ('get_username', 'payment', 'customer_name', 'comment', 'member_guess', 
                'timestamp', 'transaction_id', 'get_amount_display', 'approval')
```

stregsystem/admin.py
```diff
@@ -244,10 +250,41 @@ def get_amount_display(self, obj):
...
+        if obj.member is not None:
+            return obj.member.username
+        else:
+            return "Not assigned"
+
+    get_username.short_description = "Payment assigned to"
+    get_username.admin_order_field = "member__username"
+
+    def get_amount_display(self, obj):
+        if obj.amount is None:
```
BoAnd, line 275: Better calling this function. So we have this code one place https://github.com/f-klubben/stregsystemet/blob/9104dae9b620d424d150a73ba57409939964bc52/stregsystem/templatetags/stregsystem_extras.py#L8

stregsystem/views.py
```diff
@@ -288,3 +292,128 @@ def batch_payment(request):
...
     })
+
+
+logger = logging.getLogger('django')
+
+
+@staff_member_required()
+def import_mobilepay_csv(request):
+    def parse_csv_and_create_mbpayments(csvfile):
+        fembers = Member.objects.all()
```
BoAnd, line 303: Maybe only get active members.

BoAnd (commented):
Small thing. I have not test run the code :'(

stregsystem/admin.py
cogitantium: Absolutely, this was copy-paste reuse from `PaymentAdmin`. Both are fixed in ece3c41.

stregsystem/admin.py
cogitantium: Fixed in ece3c41. Note that this PR is still draft and not entirely done :smile:

stregsystem/views.py
cogitantium: Hmm. On production we've got around 3700 members. I'm not sure what the ratio is between active/inactive members.

What about the case if a member is (maybe by mistake) inactive, makes a mobilepayment, and only after processing gets their account made active?

For comparison, a csv with ~450 transaction _only_ takes 7 seconds in SQL execution time from start to finish.

stregsystem/admin.py
BoAnd: Hi hi

stregsystem/views.py
BoAnd: If I remember correct. Inactive members are members where the mail has bounce or some other member has take over the username. So the biggest problem is that the username maybe is not unique and how should we handle this. 

P.s maybe all usernames are unique. I would check this if ITS had open port 80. A** H***

hrjakobsen: I'm pretty sure that usernames are not unique. I added checks when creating/updating members, but those checks has  not been there for long.

stregsystem/models.py
```diff
@@ -324,6 +327,25 @@ def delete(self, *args, **kwargs):
...
+    payment = models.OneToOneField(Payment, on_delete=models.CASCADE, null=True, blank=True)
+    customer_name = models.CharField(max_length=64)
+    timestamp = models.DateTimeField()
+    amount = models.IntegerField()
+    # mobilepay transaction ids are no longer than 17 chars and assumed to be unique
+    transaction_id = models.CharField(max_length=32, unique=True)
+    comment = models.CharField(max_length=128)
+    member_guess = models.ForeignKey(Member, on_delete=models.CASCADE, null=True,
+                                     blank=True, related_name='member_guess')
+    approval = models.BooleanField(default=False)
```
VirtualSatai, line 342: Overvej at kalde det "approved" hvis det er et flag. Alternativt så gem navnet/brugeren på medlemmet af TREOen som udførte godkendelsen, (og null hvis ikke godkendt). Eller får man dette fra `payment` objektet?

stregsystem/views.py
```diff
@@ -288,3 +292,115 @@ def batch_payment(request):
...
+            mobile_payment = MobilePayment(member=None, amount=row[2].replace(',', ''),
+                                           timestamp=datetime.datetime.fromisoformat(row[3]),
+                                           customer_name=row[4], transaction_id=row[7], comment=row[6])
+            try:
+                # unique constraint on transaction_id must hold before saving new object
+                mobile_payment.validate_unique()
+
+                # do exact case sensitive match
+                match = Member.objects.filter(username__exact=mobile_payment.comment.strip())
+                if match and match.count() == 1:
```
VirtualSatai, line 323: Jeg er ikke sikker på at case sensitive match er en god ide. Brugernavne er normalt case insensitive. F.eks. er køb ikke case sensitive: https://github.com/f-klubben/stregsystemet/blob/56adc6b12c9ff85f6fa4727d0d3c7177523767d9/stregsystem/views.py#L85

Det er nok også en god idé at tilføje "active=True". Det er nok undtagelsen at folk er inaktive og indbetaler via mobilepay, så må TREOen håndtere det manuelt.

VirtualSatai (approved):
God ide at få dette lavet. @DelusionalLogic  og jeg havde tidligere tænkt på at lave det samme, men ifølge daværende formand kunne man ende i problemer hvis ikke der var en person i loop'et. Denne løsning har ikke samme problem, så go for it :)

stregsystem/views.py
cogitantium: Correct, usernames are __historically__ not unique. I just checked on production and it turns out that there's around 500 members that has duplicate usernames, most of these users are likely old `*@cs.aau.dk` members. 
However, there's no active members that has duplicate usernames. The fix you introduced in #178 should keep it this way.

If we restrict `MobilePayment` creation to only active members, we can assume that usernames are unique.
At any rate, there should be a check for duplicate usernames, I think.

stregsystem/views.py
cogitantium: Enig med case insensitivity, jeg troede faktisk at usernames var case sensitive.

Mht. active users endte jeg på denne løsning: https://github.com/f-klubben/stregsystemet/pull/204#discussion_r487716083

stregsystem/models.py
cogitantium: Renaming til 'approved' virker fornuftigt, hvis det bare skal være et flag. 
Jeg er ikke nået til testing om hvorvidt admin user bliver associeret med en given Payment. Så vidt jeg husker, var det et problem med 'batch-payment' featuren #140 som blev fikset i #198, så det er muligt og skal bare gøres.

stregsystem/views.py
cogitantium: d9ccedb: Fixed by restricting member matching to only active users.

stregsystem/views.py
cogitantium: Fixed i d7ce413

stregsystem/models.py
cogitantium: 9085e22: begge :fixd: således at et approved MobilePayment associeres med en admin user, `approval` renamed til `approved`. Jeg valgte at beholde flaget, på trods af at referencen til given admin user egentlig udtrykker det samme, da det gør formset-generation nemmere.

cogitantium:
I think the PR is about ready for real review. Some of the approaches I've chosen may not be very good, please do critique it as this is the first real Django I've written.

The UI could probably use a guiding hand from an IxD'er. It is, however, fully functional in its current state: 
image

hrjakobsen:
A few first impressions:

image

Member guess, customer name, comment should all be readonly, and member guess should not be a drop down (will increase load time and decrease performance with many members).

Ignored & Approved should be radio buttons (maybe with a third option "Skip" to simulate the current behaviour of not selecting either of the two).

All of this can be handled by creating a custom django form(set) instead of using modelformset_factory

stregsystem/models.py
```diff
@@ -324,6 +330,67 @@ def delete(self, *args, **kwargs):
             super(Payment, self).delete(*args, **kwargs)
 
 
+class MobilePayment(models.Model):
+    member = models.ForeignKey(Member, on_delete=models.CASCADE, null=True,
+                               blank=True)  # nullable as mobile payment may not have match yet
+    payment = models.OneToOneField(Payment, on_delete=models.CASCADE, null=True, blank=True,
```
joandrsn, line 336: I feel like this field does not have a high coherency with the function of this object. It gives a higher coupling with the rest of the system by having a reference to what payment was created by this object. While this may be relevant in an historical sense, I think increases the complexity of the system.
Personally, I would prefer using log entries to do investigation into the object created/edited.
But, I will also leave the design decision up to you.

stregsystem/models.py
```diff
@@ -324,6 +330,67 @@ def delete(self, *args, **kwargs):
             super(Payment, self).delete(*args, **kwargs)
 
 
+class MobilePayment(models.Model):
+    member = models.ForeignKey(Member, on_delete=models.CASCADE, null=True,
```
joandrsn, line 334: Is there a reason for differentiating between `member` and `member_guess`? I feel like this could have been a single field.

stregsystem/models.py
```diff
@@ -324,6 +330,67 @@ def delete(self, *args, **kwargs):
...
+    customer_name = models.CharField(max_length=64)
+    timestamp = models.DateTimeField()
+    amount = models.IntegerField()
+    transaction_id = models.CharField(max_length=32,
+                                      unique=True)  # trans_ids are at most 17 chars, assumed to be unique
+    comment = models.CharField(max_length=128)
+    member_guess = models.ForeignKey(Member, on_delete=models.CASCADE, null=True,
+                                     blank=True, related_name='member_guess')
+    approved = models.BooleanField(default=False)
+    ignored = models.BooleanField(default=False)
```
joandrsn, line 347: These could be chosen to be a [choice](https://docs.djangoproject.com/en/2.2/ref/models/fields/#choices). 
That could also ignore the possibility for the user to choose both "Approve" and "Ignore" on the same object.

stregsystem/models.py
```diff
@@ -324,6 +330,67 @@ def delete(self, *args, **kwargs):
             super(Payment, self).delete(*args, **kwargs)
 
 
+class MobilePayment(models.Model):
+    member = models.ForeignKey(Member, on_delete=models.CASCADE, null=True,
+                               blank=True)  # nullable as mobile payment may not have match yet
+    payment = models.OneToOneField(Payment, on_delete=models.CASCADE, null=True, blank=True,
+                                   unique=True)  # Django does not consider null == null, so this works
+    customer_name = models.CharField(max_length=64)
```
joandrsn, line 338: Hmm, there is a difference between what the CSV provides, compared to the [MobilePay Transaction Reporting API](https://github.com/MobilePayDev/MobilePay-TransactionReporting-API#transactions-endpoint). 

The matching of customer name's is not explained on the Fiki - I think that this could be more confusing that helpful for the user of the system.

requirements.txt
```diff
@@ -6,3 +6,4 @@ regex==2017.07.28
 git+https://github.com/spulec/freezegun.git@39a0f1c103e734e1f88a788e956e7c9fcffb7918
 Django-Select2==5.11.1
 django-debug-toolbar==1.8
+backports-datetime-fromisoformat==1.0.0 # required to use datetime.fromisoformat in <py37
```
joandrsn, line 9: I have yet to examine it, but a potential solution for this problem could maybe be to use Django util's `parse_datetime` which can be found here: https://docs.djangoproject.com/en/2.2/_modules/django/utils/dateparse/

joandrsn (commented):
There are quite a lot of style changes in the files. I would suggest reverting these in order to focus this pull-request on specifically on the import and processing of MobilePay transactions.

Also I don't know how I feel about the `reload-django test.sh` script. Was this meant to be committed?

stregsystem/models.py
cogitantium: As discussed in person, I agree that the function which the payment foreign key gives, could be emulated by using log-entries. However, I believe that having this reference and the options set; unique, but nullable, expresses exactly the behaviour we wish a MobilePayment to have. Instead of handling this manually through log-entries, we can simply use the Django ORM to enforce these constraints for us.
My vote is to keep it as a foreign key.

stregsystem/models.py
cogitantium: Yes, the member-field is the member for which the MobilePayment will add a Payment for. The member-guess-field _may_ be the member that the customer meant when writing their comment. It's important to differ between suggestions and exact matches, so that a Treo does not blindly approve suggestions.

stregsystem/models.py
cogitantium: I agree, this becomes possible to create nice UI for when I make the custom Django Form.

cogitantium: The choices could be something like; `{Unprocessed, Approved, Ignored}`.

stregsystem/models.py
cogitantium: Oh no, I hoped it would contain that. In that case, eventually API-imported MobilePayments can have their customer_name set to empty. This currently on affects the member_guess function.

requirements.txt
cogitantium: Good idea, fixed in 53bdc3c and 40a876a

joandrsn (approved):
LGTM

stregsystem/models.py
```diff
@@ -309,10 +313,10 @@ def save(self, *args, **kwargs):
             # TODO: Make atomic
             self.member.make_payment(self.amount)
             super(Payment, self).save(*args, **kwargs)
-            self.member.save() 
+            self.member.save()
             if self.member.email != "":
                 if '@' in parseaddr(self.member.email)[1] and self.member.want_spam:
-                    send_payment_mail(self.member, self.amount)            
+                    send_payment_mail(self.member, self.amount)
```
hrjakobsen, line 319: Some formatting changes here which are not part of the PR, but I'll allow it :)

cogitantium: Damn, thought I reverted them all!

cogitantium:
I feel like this feature is in a good place for proper reviews now. 

Suggestions are very much welcome from the old fitters; @hrjakobsen, @VirtualSatai, @DelusionalLogic.

stregsystem/fixtures/mobilepay-sales-testdata-fixture.csv
```diff
@@ -0,0 +1,7 @@
+Event;Currency;Amount;Date and time;Customer name;MP-number;Comment;TransactionID;Payment point;MyShop-Number;Date;Time
```
GeneralDenmark, line 1: Super fedt med fixtures! :+1:

stregsystem/migrations/0011_mobilepaytool_20201027_1623.py
```diff
@@ -0,0 +1,29 @@
...
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('customer_name', models.CharField(max_length=64)),
+                ('timestamp', models.DateTimeField()),
+                ('amount', models.IntegerField()),
+                ('transaction_id', models.CharField(max_length=32, unique=True)),
+                ('comment', models.CharField(max_length=128)),
+                ('status', models.CharField(choices=[('U', 'Unset'), ('A', 'Approved'), ('I', 'Ignored')], default='U', max_length=1)),
+                ('member', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to='stregsystem.Member')),
+                ('member_guess', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='member_guess', to='stregsystem.Member')),
```
GeneralDenmark, line 25: Er der nogen specifik grund til at vi gemmer member_guess i databasen?

stregsystem/templates/admin/stregsystem/mobilepaytool.html
```diff
@@ -0,0 +1,142 @@
...
+                                    {% elif field.name == "member_guess" %}
+                                        {{ field.value|get_member_str_key|safe }}
+                                        <span style="visibility: hidden; position: absolute">{{ field }}</span>
+                                    {% elif field.name == "customer_name" %}
+                                        {{ field.value }}
+                                        <span style="visibility: hidden; position: absolute">{{ field }}</span>
+                                    {% elif field.name == "comment" %}
+                                        {{ field.value }}
+                                        <span style="visibility: hidden; position: absolute">{{ field }}</span>
+                                    {% elif field.name == "status" %}
```
GeneralDenmark, line 128: `style="visibility: hidden; position: absolute" `
would be better as a class imo
`.hidden {
        visibility: hidden;
        position: absolute;
}`

stregsystem/tests.py
```diff
@@ -1073,7 +1075,7 @@ def test_creates_warning_for_duplicate_usernames(self):
         response = self.client.post(reverse('admin:stregsystem_member_change', kwargs={'object_id':2}), model_to_dict(self.jeff2), follow=False)
 
         messages = list(get_messages(response.wsgi_request))
-
```
GeneralDenmark, line 1076: ?

stregsystem/tests.py
```diff
@@ -1085,7 +1087,7 @@ def test_no_warning_unique_usernames(self):
...
```
GeneralDenmark, line 1088: !

GeneralDenmark (approved):
All in all I think this is great :)

stregsystem/templates/admin/stregsystem/mobilepaytool.html
cogitantium: God idé! Nice fix i 6c4ab2e

stregsystem/migrations/0011_mobilepaytool_20201027_1623.py
cogitantium: Hm, nej - faktisk ikke, udover at det er praktisk at have det behandlet ved import og let at trække ud i modelformset. 

Jeg har lige testet og kan se at load-time er ca. dobbelt så langsom når den er med, da den stadig skaber dropdown-widget selvom det er hidden.

Det bør nok fikses inden produktion.

stregsystem/tests.py
cogitantium: :upside_down_face:

stregsystem/tests.py
cogitantium: :upside_down_face:

stregsystem/admin.py
```diff
@@ -236,9 +238,44 @@ def get_username(self, obj):
...
+
+
+class MobilePaymentAdmin(admin.ModelAdmin):
+    list_display = (
+        'get_username', 'payment', 'customer_name', 'comment', 'member_guess', 'timestamp', 'transaction_id',
+        'get_amount_display', 'status')
+    valid_lookups = 'member'
+    search_fields = ['member__username']
+    autocomplete_fields = ['member']
+
```
hrjakobsen, line 261: you should add `raw_id_fields = ['member_guess']` so that it doesn't generate a dropdown of all fembers

cogitantium: Sound idea, thanks; 8ebb003

stregsystem/migrations/0011_mobilepaytool_20201027_1623.py
cogitantium: @hrjakobsen havde en fiks idé og har forbedret performance af member_guess i modelform væsentligt. Så det er ikke længere et væsentligt problem i mine øjne.

Der er selvfølgelig argumentet om det er spild af plads at lagre det i database, men da det kun er en foreign key, er jeg okay med det.

Hvorvidt det kan være problematisk at lave beregne member_guess ved import-tid eller form-generations-tid er minimal i mine øjne. Hvis member ikke eksisterer ved import-tid er det højest sandsynlig en tilmelding, som alligevel skal behandles manuelt og ikke supporteres af dette tool. 
Værste fald er at Treoen skal vælge memberen ved fuzzy-search i de få tilfælde det kan være gældende.

GeneralDenmark (approved):
Yes sir!

stregsystem/migrations/0011_mobilepaytool_20201027_1623.py
GeneralDenmark: Du har ret min ven!

cogitantium:
Mobilepay API import of transactions now works and is integrated with MobilePaytool. I think the interface is pretty much finished.

image

DelusionalLogic (approved):
I feel better living in a world where this code is inside the stregsystem.

stregsystem/models.py
```diff
@@ -324,6 +329,81 @@ def delete(self, *args, **kwargs):
...
+            if processed_payment.status == MobilePayment.APPROVED:
+                payment = Payment(
+                    member=processed_payment.member,
+                    amount=processed_payment.amount)
+            elif processed_payment.status == MobilePayment.IGNORED:
+                payment = Payment(
+                    member=processed_payment.member,
+                    amount=0)
+            else:
+                raise RuntimeError("Trying to process MobilePayment not of status (APPROVED, IGNORED)")
```
GeneralDenmark, line 382:
```suggestion
           else:
                payment = Payment(
                    member=processed_payment.member,
                    amount=0)
```
the function make_processed_mobilepayment_query makes this check. It can only return MobilePayment.IGNORED and MobilePayment.APPROVED :slightly_smiling_face:

stregsystem/templates/admin/stregsystem/mobilepaytool.html
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        .submit-button {
+            min-width: 180px;
+        }
+
+        #submit-matched-button {
+            background-color: #E6AE07;
+        }
+    </style>
```
GeneralDenmark, line 45: I like that you are using css classes, though, with this many, you might consider throwing them into a .css file, and importing them ind the header :+1:

stregsystem/templates/admin/stregsystem/mobilepaytool.html
```diff
@@ -0,0 +1,191 @@
...
+                        </tr>
+                    {% endfor %}
+                </table>
+                <br>
+                <hr style="height: 3px">
+                <br>
+                <input type="submit" class="submit-button" id="submit-matched-button" name="action"
+                       onclick="return prompt()"
+                       value="Submit matched payments"/>
+                <span style="color: white">fuck css</span>
```
GeneralDenmark, line 185: jeg er sikker på at der er en måde at "snyde" darkreader på, så den fortsat er skjult.
måske give den klassen: '--darkreader-text-inverted'?

stregsystem/views.py
```diff
@@ -288,3 +292,64 @@ def batch_payment(request):
         "select2_js": settings.SELECT2_JS,
         "select2_css": settings.SELECT2_CSS,
     })
+
+
+@staff_member_required()
+def mobilepaytool(request):
+    paytool_form_set = modelformset_factory(MobilePayment, form=MobilePayToolForm, extra=0, fields=(
+        'timestamp', 'amount', 'member', 'comment', 'status'))  # TODO: 'customer_name' removed,
+    #  MobilepayAPI does not that information at this point in time - add back 'customer_name' if available in future
```
GeneralDenmark, line 301:
```suggestion
    #  MobilepayAPI does not have that information at this point in time - add back 'customer_name' if available in future
```

GeneralDenmark (approved):
Awesome to have implemented logging, though I think this maybe should've been its own PR, yet all other comments I have are minor things.

Good job, with everything! :+1: :)

stregsystem/templates/admin/stregsystem/mobilepaytool.html
cogitantium: Could do, setting class to "--darkreader-text-invert" didn't trigger Darkreader on my machine though.

stregsystem/models.py
cogitantium: Good point, this check is just redundant :+1:

stregsystem/templates/admin/stregsystem/mobilepaytool.html
cogitantium: Absolutely, fixed in f6e428e

stregsystem/templates/admin/stregsystem/mobilepaytool.html
GeneralDenmark: Tbh, i didn't check, as Qutebrowser's greasymonkey support for Darkreader is limited 👍 But this is not really an issue, just more of a "thought" if you will 😄 

Beside I highly doubt that there is that many users of the admin tool for Stregsystemet, yet even fewer that has a css mod like Darkreader.

cogitantium: Good point. Also, if you want your easteregg to be found - you shouldn't hide it too well :)

krestenlaust:
VippsMobilepay-fusion's new API might have opened up for automatic matching without having to write information in the comment-feature on transactions. The last 4-digits and legal name of payer is received. #416 Documents implementation of this new API.

PR #207, opened 2020-09-26
Added ability to view logentries as admin

### LINKED ISSUE #206: GUI for Admin LogEntry
It could be nice to have some GUI for the admin logentries. 

Link to an implementation: https://medium.com/datadriveninvestor/monitoring-user-actions-with-logentry-in-django-admin-8c9fbaa3f442

joandrsn:
Fixes #206 

Adds the ability to see the logs like this:
image

However, this also adds it to the top of the applist.
image

I see this as a good thing as we then can see who makes balance changes if we write to the LogEntry.

### DIFF
--- a/stregsystem/admin.py
+++ b/stregsystem/admin.py
@@ -2,6 +2,7 @@ from django.contrib import admin
 from django import forms
 from django.contrib.admin.views.autocomplete import AutocompleteJsonView
 from django.contrib import messages
+from django.contrib.admin.models import LogEntry
 
 from stregsystem.models import (
     Category,
@@ -243,7 +244,27 @@ class PaymentAdmin(admin.ModelAdmin):
     get_amount_display.short_description = "Amount"
     get_amount_display.admin_order_field = "amount"
 
+class LogEntryAdmin(admin.ModelAdmin):
+    date_hierarchy = 'action_time'
+    list_filter = ['user', 'content_type', 'action_flag']
+    search_fields = ['object_repr', 'change_message']
+    list_display = ['action_time', 'user', 'content_type', 'object_id', 'action_flag', 'change_message']
 
+    def has_view_permission(self, request, obj=None):
+        return True
+
+    def has_add_permission(self, request):
+        return False
+
+    def has_change_permission(self, request, obj=None):
+        return False
+    
+    def has_delete_permission(self, request, obj=None):
+        return False
+
+    
+
+admin.site.register(LogEntry, LogEntryAdmin)
 admin.site.register(Sale, SaleAdmin)
 admin.site.register(Member, MemberAdmin)
 admin.site.register(Payment, PaymentAdmin)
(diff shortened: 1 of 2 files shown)


hrjakobsen:
Can this be seen by all staff or only superusers?

joandrsn:
Currently, its all staff. Should I make it superusers only? - Do we even have superusers?

hrjakobsen:
We do have super users, and we also have users who only has very restricted access (bread razzia users for example). We should have some way of controlling who can access it. Can the default Django permissions not do this? I recall setting permissions for individual classes before.

joandrsn:
My bad @hrjakobsen.
The permissions are controlled by the default permissions. The "Logmeddelelser" is invisible to regular users.
I created a test user without any permissions to see what would happen:
image

The user list has the following permissions:
image
Since I have the functions `has_add_permission`, `has_change_permission` & `has_delete_permission` returning `False` and only `has_view_permission` returning `True`, even when the user has all the permission for add/edit/delete, he/she cannot add/edit/delete a LogEntry - only view.
Although, as soon as the user has a add/edit/delete permission, he/she also gains the ability to view the logentries.

cogitantium (approved):
Nice, this ties in well with #204

stregsystem/admin.py
```diff
@@ -243,7 +244,27 @@ def get_amount_display(self, obj):
     get_amount_display.short_description = "Amount"
     get_amount_display.admin_order_field = "amount"
 
+class LogEntryAdmin(admin.ModelAdmin):
+    date_hierarchy = 'action_time'
+    list_filter = ['user', 'content_type', 'action_flag']
+    search_fields = ['object_repr', 'change_message']
```
hrjakobsen, line 250: I think you should be able to search for the user rather than choosing it in the filter list.

joandrsn: Is this because the number of users could be large? Unfortunately, there is only 1 search field on the list. I'm thinking that searching for a certain users changes could be kinda difficult.

hrjakobsen: Yes the user list could become large

Would adding `'user__username'` to `search_fields` not do the trick? I can't remember whether you can search for nested fields like that.

joandrsn: I just tested, and I believe you can.

hrjakobsen (approved):
I don't know if we should add object_repr to the list view. Maybe it will take up too much space, but it would be nice to see on which members actions were performed.

No matter if a change is made to reflect this, i approve the PR.

joandrsn:
Hmm, that may be a good idea @hrjakobsen.

I just added it to my local repo and here's how it looks:
image

Maybe @cogitantium can also say if he likes it?

joandrsn:
I talked with @cogitantium and he approves of the added `object_repr`.
He did however find a potential security bug.
While a regular user would not have the "Logmeddelelser" on the overview, if the user found the url `<URL>/admin/admin/logentry/` he would still be able to see the logentries. This has been fixed and now returns a `403 Forbidden`-message if the user is not superuser.

cogitantium (approved):
Just for good measure, I approve the latest changes :+1: 

I think restricting global log entry history to superusers is wise.

PR #209, opened 2020-10-01
Feature timezone aware userinfo

### LINKED ISSUE #208: Dates on user information's last 10 purchases are wrongly displayed
This is a screenshot from my personal purchase history.
image

The purchases are correct, but I believe they are wrongly displayed.
I did not purchase anything at 6:30 in the morning, but rather around 8:30. I think that it is because timestamps in the databased are stored in UTC while I as a user would expect it to displayed with my local timezone (Currently CEST).

cogitantium:
Fixes #208 by writing timestamp for a user on userinfo in Europe/Copenhagen. Works fine from preliminary tests but unfortunately assumes the purchasing user is located in Denmark - sorry @hrjakobsen.


joandrsn (approved):

hrjakobsen (approved):
We should consider setting the default timezone in `settings.py` to "Europe/Copenhagen" instead of the current "UTC". The timestamps will still be stored in UTC on the server.

cogitantium:
Closed by #211

PR #210, opened 2020-10-02
Initial support for Custom ID's on Products

### LINKED ISSUE #35: Migrate Purchase code from database id to one or more custom IDs
Previously discussed in rewrites to stregsystemet is to change the purchasing code of a product from it's database id to one or more custom but still unique IDs.

Think of them as SKU. Right now all products have a single unique SKU which is it's database id. Moving forward it would be nice if the purchase code or SKU could be changed or customized separately from bumping the database ID. F-FLAN could have 1337 then e.g. or F-LAN if we allowed alphanumeric characters.

A simple move to this should be possible simply expanding the Product model to have a relation to a SKU model and during migration simply copy each products current ID to the SKU meaning that all products would have the same purchasing code. Leaving us with the option to change or add new SKUs for products.

Would be a great feature and would still maintain old relations in sales as no actual changes are made - only where the code people type is coming from.

joandrsn:
I have added a new field (`custom_id`) to Product class.
Using migrations, this `custom_id` will be filled with the primary key from the Products table.
It must be unique, not null and not blank.

I have added the `custom_id` to tests to ensure that all test data have it filled. 

The custom id is currently only an integer.
Fixes #35


VirtualSatai (approved):
I like this change as a first step towards custom ids. It could be considered to make this custom_id a string instead of a number, that way, only minimal changes to the buy-string parser would be needed (if any?), to support alphabetical id's (like "flan").

cogitantium:
If we're going to alter the product ids used for buying, I second @VirtualSatai suggestion of allowing alphanumerics - let's go all the way, if we're doing it.

I think allowing alphanums, requires us to revise the EBNF and purchase-parser and introduce restricted keywords for custom_ids; e.g. usernames.

PR #213, opened 2020-10-05
More bats

AndersSpringborg:
As the system now runs on mobile phones, the client do not crash with an increase of bats on the screen. Trippling the amount of bats, seems reasonable

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ For Ubuntu with virtual envs:
 1. Install python3 with pip
  - `sudo apt install python3 python3-pip`
 2. Create virtual environment
- - `python3 -m vevn venv`
+ - `python3 -m venv venv`
 3. Activate virtualenv
  - `source venv/bin/activate`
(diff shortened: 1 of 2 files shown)


hrjakobsen:
Let's spawn 2^n bats instead

AndersSpringborg:
agree

BoAnd:
Hey guys. 

The only correct algorithm to this problem. Would be the Ackermann function.
Where A(d.getMonth(), d.getDate())

hrjakobsen:
My brain hurts just looking at the definition of that function 

Ackermann function on Wikipedia

cogitantium (commented):
I tried to run it with 2^30 bats but even though I [downloaded more ram](https://downloadram.net/) it still hung my machine. Maybe a larger number of bats would help?

AndersSpringborg:
Maybe i would help with an extra graphic, so the bats would't be so lonely

PR #214, opened 2020-10-08
Bugfix escape username

joandrsn:
The `sales_hint` was unescaped, resulting in the following hint:
image

After this fix, it should be presented as this:
image

If this is accepted, please squash the commit since I made some indentation errors :)

### DIFF
--- a/stregsystem/templates/stregsystem/menu.html
+++ b/stregsystem/templates/stregsystem/menu.html
@@ -107,16 +107,18 @@ har {{member.balance|money}} kroner til gode!
 {% else %}
 <p>Ingen produkter.</p>
 {% endif %}
+{% endautoescape %}
+{% endblock %}
 </center>
 <br />
 {% if give_multibuy_hint %}
     <center>
         <h1>psssst. multibuy er enabled.</h1>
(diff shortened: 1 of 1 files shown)


stregsystem/templates/stregsystem/menu.html
```diff
@@ -117,6 +115,8 @@ <h1>psssst. multibuy er enabled.</h1>
         <b> Du kunne have skrevet: {{sale_hints}} </b>
     </center>
 {% endif %}
+{% endautoescape %}
+{% endblock %}
```
hrjakobsen, line 119: I think it's better to keep the `products` block as it is, and start a new {% autoescape off %} above the saleshints, as the `sale_hint` is not part of the product list. and will make extending this file (if ever required) more difficult.

hrjakobsen (changes requested):

stregsystem/templates/stregsystem/menu.html
joandrsn: Thank you! I changed it.

hrjakobsen (approved):

cogitantium (approved):

Issue #188, opened 2019-10-18
Allow user to cancel latest bought item


MathiasMehl:
This could be useful when a user pays for an item, and it turns out the item is not in stock, or simply that a user wants another item instead. 

The stregsystem already trusts and assumes that the user will not cheat, so this feature should not encourage cheating the stregsystem. 

There could be a fixed time that a cancel is possible, e.g 1 min after buying an item.

hrjakobsen:
_If_ this is implemented, we have to figure out how to manage the accounting for this. This would probably entail creating a sale with a negative price, as well as adding a payment to the user, as to retain the correctness of the logged sales information.

jones-g:
I would not say I'm completely against this but usually this has been "fixed" by simply picking up the purchased product when it is in stock again.

PR #215, opened 2020-10-09
Added initial support for reimbursement

joandrsn:
This is an initial fix for #188.

This PR allows the user to refund his/her purchase by doing the following:
Modifies the sale to show `is_reimbursed`.
Creates a new payment for the amount that the sale was on. This payment is tagged with `is_reimbursement`.
The button for reimbursing the user is on the user information. The user only sees sales that have not been reimbursed. 
Example:
image
[Example of confirmation dialog](https://user-images.githubusercontent.com/5766695/95626957-9b701080-0a7b-11eb-9839-479afe469605.png)

The button is only visible to the user if the sale was within the last 12 hours.
Unfortunately I had to allow the user to edit a sale. This may be unwanted - but I could not think of another way to do this.

The reason for modifying the sale is to allow the values to be filtered out of statistics.

Questions I had during making this:
Should the "Undo" button be an actual button or rather an anchor?
Should I have instead of creating a `Payment`, instead save `Refund` in a separate table?

I hope for feedback.

Also I need tests. Lots of tests...

### DIFF
--- a/stregsystem/models.py
+++ b/stregsystem/models.py
@@ -137,6 +137,19 @@ class Order(object):
         # We changed the user balance, so save that
         self.member.save()
 
+class Reimbursement(object):
+    #Limits the amount of hours a reimbursement is available.
+    MAX_REIMBUSEMENT_HOURS = 12
+
+    @transaction.atomic
+    def from_sale(sale_id):
+        sale = Sale.objects.get(id=sale_id)
+        if not sale.is_reimbursable:
+            return
+        sale.is_reimbursed = True
+        sale.save()
+        Payment.objects.create(amount=sale.price, member=sale.member, is_reimbursement=True)
+
 
 class GetTransaction(MoneyTransaction):
     # The change to the users account
@@ -286,6 +299,8 @@ class Payment(models.Model):  # id automatisk...
     member = models.ForeignKey(Member, on_delete=models.CASCADE)
     timestamp = models.DateTimeField(auto_now_add=True)
     amount = models.IntegerField()  # penge, oere...
+    is_reimbursement = models.BooleanField(default=False, null=False,
+                                            help_text='Viser om en betaling er tilbageført')
 
     @deprecated
     def amount_display(self):
@@ -425,6 +440,8 @@ class Sale(models.Model):
     room = models.ForeignKey(Room, on_delete=models.CASCADE, null=True)
     timestamp = models.DateTimeField(auto_now_add=True)
     price = models.IntegerField()
+    is_reimbursed = models.BooleanField(default=False, null=False, 
+                                        help_text='Viser om et salg er tilbageført',)
 
     class Meta:
         index_together = [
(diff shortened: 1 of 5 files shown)


VirtualSatai:
I like the idea, though I think it would need the approval/blessting of the TREO/Forstyrrelse. For modelling the data, I think it would be better to have the reference between the Payment and the (now refunded) Sale, rather than having two boolean flags. 

This way, if you want to query for refunded sales, you can simply check for non-null values, and the same for the negative Payments, which is the refund.

cogitantium:
> I think it would need the approval/blessting of the TREO/Forstyrrelse

This should definitely be possible given a sensible implementation, in my opinion. I've begun the discussion for the relevant parties.

stregsystem/models.py
```diff
@@ -137,6 +137,19 @@ def execute(self):
...
+    MAX_REIMBUSEMENT_HOURS = 12
+
+    @transaction.atomic
+    def from_sale(sale_id):
+        sale = Sale.objects.get(id=sale_id)
+        if not sale.is_reimbursable:
+            return
+        sale.is_reimbursed = True
+        sale.save()
+        Payment.objects.create(amount=sale.price, member=sale.member, is_reimbursement=True)
```
hrjakobsen, line 151: We should add a  [LogEntry](https://stackoverflow.com/a/7905253) to the Payment object to note that it is a result of a refund.

joandrsn: While I think this is a great idea, during implementation, I found out that LogEntry has a `NOT NULL` constraint on the `user_id`. So that pretty much rules it out.
https://github.com/django/django/blob/b9fe7f9294b1b4fc974c008adeb96e1375cdb0c6/django/contrib/admin/models.py#L45

An idea would be to make a "reimbursement user" and set that user as the person who did it. Buut that is kinda hacky.

Any ideas on how to overcome this?

joandrsn:
@VirtualSatai I thought about doing a foreign key to the refunded sale, but I believe that would make the list of sales slow.
Let's say we wanted to show a list of sales, but we would like to exclude all of the sales that were refunded. The only way for a sale to know if it was reimbursed is if there is a payment which have a foreign key to it.

The double boolean solution may be simple, but since the information is highly local, I think this is still the best solution at the moment. If you still think otherwise, then please enlighten me. I cannot think of the usecase where it would be better to have a foreign key over the two booleans.
Maybe only if in a data analysis, that the user wanted to be 100 % sure that a reimbursed sale was related to a payment. But what would that information be useful for?



Anyway, I have made 3 tests, testing the reimbursement feature and some changes.
I think the PR is ready for review.

stregsystem/models.py
Zinoex: I would actually prefer if the person who did the reimbursement was referenced on the LogEntry. That allows for tracability of reimbursements. The Payment object to be reimbursed should also be referenced.

joandrsn: Completely agree, but since stregsystemet has modeled it's own members seperate from django users, it does not have users and thus it seems impossible to use the LogEntry table since LogEntry has a `NOT NULL` constraint to users.

falkecarlsen:
> With the implementation of inventory management, it is key that we add +1 to the quantity of the reimbursed product, if the product quantity has a start date. If we do not increment it, we do not reflect the actual inventory status, which can lead to the system not allowing you to buy a given product (see #274)

I'm fairly certain that deleting a sale object makes quantity of given product increment, furthermore, the purchase price is returned to the purchaser. If that's the case, we don't need to do very much, except allow that action for the users within the grace period.

PR #216, opened 2020-10-10
Adds support for videos to be displayed in the kiosk

### LINKED ISSUE #147: Allow video in the Kiosk
To allow for videos like this: https://streamable.com/l3nfp

GeneralDenmark:
Fixes #147

This PR changes the table kioskitem to have a django file input field instead of an image field, as image field is inheriting from input field, and we are not using the specific parameters for image fields, the conversion between these fields are seamless.

For the frontend to support a video, we just needed to inject the video element into the dummy div.
 
As the ImageField had its own validation regarding what file extensions that it could take (only valid image files), I had to create a custom validator for the filed, to avoid people uploading none-image and none-movie files.


cogitantium (changes requested):
Nice, great to finally have this feature! A few things:

1. The model field should reflect that is is media and not just images in admin like this. I'm not sure how to handle this migration nicely as it requires setting defaults for existing kiosk items:
image

1. When viewing media in a more vertical format than 16:9, the F-Klubben splash image is visible in the background. It seems like any media is simply displayed above the splash image:
image

GeneralDenmark:
When you are writing the reviews, please use the "request change" feature to make change requests, you've done that for two change requests (even though they were comments on wrong part of the code). It makes it easier for the one having to make the changes, to see the progress.

And when completing the requested changes, its possible to mark them as resolved :) 

else, good review, the changes you've requested have been completed.

stregsystem/migrations/0011_auto_20201011_1819.py
```diff
@@ -0,0 +1,18 @@
...
+        ('stregsystem', '0010_auto_20200305_0917'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='member',
+            name='year',
+            field=models.CharField(default='2020', max_length=4),
+        ),
+    ]
```
cogitantium, line 18: Do we want to include this auto-migration with this PR?

kiosk/models.py
```diff
@@ -12,5 +13,11 @@ class KioskItem(models.Model):
     notes = models.CharField(max_length=2000, blank=True, null=True)
     uploaded_date = models.DateField(auto_now_add=True)
     active = models.BooleanField(default=True)
-    image = models.ImageField(upload_to='kiosk', null=False)
+    image = models.FileField(upload_to='kiosk', null=False, validators=[validate_file_extension])
+    media = models.FileField(upload_to='kiosk', null=False, validators=[validate_file_extension])
```
cogitantium, line 17: Is it possible to update (with similar tooling as you've written in kiosk/migrations/0006_auto_20201010_1459.py ) the historical `image` kiosk-entries to become `media` kiosk-entries, such that we only have one filefield; `media`?

I think someone more competent at Django than me should have a look at this.

cogitantium (commented):
Looks good, I still have some thoughts on the migrations and I'm not certain on which approach would be best. 

I think @hrjakobsen, @VirtualSatai, and/or @DelusionalLogic may be able to give some input.

stregsystem/migrations/0011_auto_20201011_1819.py
hrjakobsen: No, it should be a PR in itself (or directly committed to next i guess)

kiosk/models.py
GeneralDenmark: It can be removed, I wrote a migration script for moving all the data from image to media. The fact that it is still in the PR, is a mistake on my part. I'll remove that posthaste.

stregsystem/migrations/0011_auto_20201011_1819.py
GeneralDenmark: Yes, this can be removed, I caught it in the first commit, but must've missed it in my second.

GeneralDenmark:
As the two previous change requests was blatant mistakes on my part, I removed them from history. Trump would be proud.

cogitantium (approved):
LGTM

DelusionalLogic (approved):
Showing the splash screen seems janky. I love it!

Edit: It's actually not always the splash screen. The way we flip the kiosk images without flickering is by having 2 elements. One in front and one in back. The one in the back has the next image so it has time to load.

To put it simply, you will have the next image behind the video if it doesn't fill the entire screen. I think this was also a problem with images though, so who cares.

GeneralDenmark:
> Showing the splash screen seems janky. I love it!
> 
> Edit: It's actually not always the splash screen. The way we flip the kiosk images without flickering is by having 2 elements. One in front and one in back. The one in the back has the next image so it has time to load.
> 
> To put it simply, you will have the next image behind the video if it doesn't fill the entire screen. I think this was also a problem with images though, so who cares.

@DelusionalLogic, I think this PR also fixes that problem, as I make sure to clear the background image and my bowel system before changing background image.


@cogitantium Thanks for the changes, although I don't really agree on the necessity for renaming all of the migrations, its not a problem, and if you wanna do it, :gunhands: you are more than whale-cum!

PR #221, opened 2020-10-24
Dangling migrations

GeneralDenmark:
This small PR avoids dangling migrations each year, by instead of setting the default value to the result of the method timezone.now().year. Its a reference to a function called get_current_year.

This method is then called when creating a new instance of the Member model.


stregsystem/models.py
```diff
@@ -148,6 +148,10 @@ def change(self):
         return self.amount
 
 
+def get_current_year():
+    return timezone.now().year
```
hrjakobsen, line 152: Returnerer det her ikke en integer, hvorimod `year` feltet forventer en string?

GeneralDenmark: Jo og nej. Implicit kald af ".__str__".. Det ville nok være bedst at gøre det eksplicit :jack_o_lantern:

cogitantium (approved):

DelusionalLogic (approved):

jenrik (approved):

PR #224, opened 2020-11-11
Fixes double spend when reloading menu_sale

jenrik:


cogitantium (approved):
Good idea with the backwards compatibility with redirect. As discussed in person, I've made login-purchases links again and wrote a stupid error-message

stregsystem/templates/stregsystem/menu.html
```diff
@@ -82,10 +82,10 @@
 			    <td>
                     <form method="post" action="/{{ room.id }}/sale/{{ member.id }}/">
                         {% csrf_token %}
+                        <a href="" onclick="this.closest('form').submit();return false;">{{product.name}}</a>
```
jenrik, line 85: I would like to state for the record that I believe a button is the superior aesthetic. But I also accept that I don't have final say in the matter since I officially left FIT

PR #225, opened 2020-11-12
Generates payment QR codes internally instead of using an external service

jenrik:


cogitantium (approved):
Nice :ok_hand: Made some minor changes;
- Restructured into util.py and forms.py
- Removed printf debug statement
- Enlargened QR codes to 300x300px

stregsystem/views.py
```diff
@@ -288,3 +290,21 @@ def batch_payment(request):
         "select2_js": settings.SELECT2_JS,
         "select2_css": settings.SELECT2_CSS,
     })
+
+
+def qr_payment(request):
+    form = QRPaymentForm(request.GET)
+    if not form.is_valid():
+        return HttpResponseBadRequest("Invalid input for MobilePay QR code generation")
```
jenrik, line 298: We should probably tell people which field they messed up or forgot

stregsystem/views.py
```diff
@@ -288,3 +290,21 @@ def batch_payment(request):
...
+        'phone': '90601',
+        'comment': form.cleaned_data.get('member')
+    }
+
+    if form.cleaned_data.get("amount") is not None:
+        query['amount'] = form.cleaned_data.get("amount")
+
+    data = 'mobilepay://send?{}'.format(urllib.parse.urlencode(query))
+
+    return qr_code(request, data)
```
cogitantium, line 310:
```suggestion
    return qr_code(data)
```

stregsystem/views.py
cogitantium: The users do not enter this information themselves. This is given by the Django template tags, e.g. `{{ data.value }}`.

cogitantium (commented):
Fixing qr_code call in views

stregsystem/views.py
jenrik: Users include current and future developers

PR #227, opened 2020-11-14
Add Codecov to Github Action

cogitantium:
I've tested Codecov against my [fork](https://github.com/cogitantium/stregsystemet/pull/4) and it seems to work well. Pricing is free for open source repos and it generates pretty graphs.

Until approval, I've held off granting Codecov access to f-klubben/stregsystemet which would be the last step in activating code coverage checks for all PRs onto `next`.

For examples of decreased coverage see [(edited) comment Codecov made here](https://github.com/cogitantium/stregsystemet/pull/4#issuecomment-727213167)


hrjakobsen:
image

I don't understand the figure it creates, but I definitely support merging this :+1:

hrjakobsen (approved):

GeneralDenmark:
> image
> 
> I don't understand the figure it creates, but I definitely support merging this

@hrjakobsen  If you click on image that the bot provides in its comment, you are redirected to the dashboard page for the stregsystem, here there are some more graphs to see, and an explanation for the graph you also posted! :slightly_smiling_face: 
[Here](https://codecov.io/gh/cogitantium/stregsystemet/pull/4)

This is a catch!

cogitantium:
image
I'm not sure how I feel about the `codecov/project` check fails if coverage decreases. Docs doesn't seem to indicate whether this behaviour can be turned off.

hrjakobsen:
Oh good point, I don't like that behaviour from codecov :-1:

hrjakobsen:
I found on their [docs](https://docs.codecov.io/docs/commit-status), that the treshold for success can be lowered. I guess a `treshold` of `100%` would be what we want.

cogitantium:
@hrjakobsen I should've paid more attention. 100% threshold seems to work as expected - and only overrides that specific option, such that defaults are kept, e.g: `require_ci_to_pass: yes`.

I think we should remove `.travis.yml` also, since we now use Github Actions to do CI and generation of coverage reports for Codecov.

GeneralDenmark:
I actually like that Codecov fails when Code coverage drops. It does not force you so that you can't merge, but it highlights that, maybe the pr author should've had more focus on writing tests.
As there are no "warnings" with the checks, the error can help to highlight that there is an issue, that the approver has to look into :100:

hrjakobsen:
To me a failing check indicates a breaking change, so I don't agree with with @GeneralDenmark in the general case. Lowering the threshold to an arbitrary value only seems to be a sanity check against (i) making a large feature without testing (ii) removing a bunch of test cases, which both seems like something a human would catch quite easily. If there was a way to indicate warning, I'd support a warning if lowering the test coverage, but in binary world I support `threshold: 100%`.

GeneralDenmark:
The problem is that, there is no way of showing a "warning", so there are either successes or errors. And in genereal, I think a reduction in tests % is _not_ an success, and thus an error..

I see your point, and would've preferred a warning instead of an error, but as there are only these two different kinds of responses. IMO an error is preferred rather than a false success.

PR #229, opened 2020-11-19
Setup Black Github Action, update tox-settings, and apply Black code style

cogitantium:
I miss having a unified code-style on stregsystemet and get annoyed when autoformatting in my IDE changes irrelevant lines.

This PR introduces the Black formatter - not chosen because I particulary prefer it but because it is pretty much unconfigurable and popular. Choosing this formatter further decreases discussions on which way is the proper way. 
Besides replacing flake8, Black also puts isort-check out of a job by doing import optimisation better.

By adding the Black Action to our Github Workflow which is triggered on pull requests onto `next`, it's clear when code are being merged which does not follow the coding style. This action does make the entire workflow fail, which feels somewhat severe - but allowing it to fail, undermines a unified code-style.


stregreport/models.py
```diff
@@ -2,24 +2,20 @@
...
-    RAZZIA_CHOICES = [
-        (BREAD, "Brødrazzia"),
-        (FOOBAR, "Foobar razzia")
-    ]
-    members = models.ManyToManyField(Member, through='RazziaEntry')
+        permissions = (("host_razzia", "Can host a foobar or bread razzia"),)
+
+    BREAD = "BR"
+    FOOBAR = "FB"
+    RAZZIA_CHOICES = [(BREAD, "Brødrazzia"), (FOOBAR, "Foobar razzia")]
```
hrjakobsen, line 12: I like the previously style better in these cases

kiosk/views.py
```diff
@@ -40,26 +37,15 @@ def find_next_media_real(request, item_id):
...
-            KioskItem.objects
-            .filter(active=True)
-            .order_by('ordering', 'id')
-            .filter(
-                Q(ordering__gt=item.ordering)
-                | (Q(ordering=item.ordering) & Q(id__gt=item.id))
-            )[0]
+            KioskItem.objects.filter(active=True)
+            .order_by("ordering", "id")
+            .filter(Q(ordering__gt=item.ordering) | (Q(ordering=item.ordering) & Q(id__gt=item.id)))[0]
```
hrjakobsen, line 42: Previous style was more readable imo.

stregreport/views.py
```diff
@@ -257,25 +255,55 @@ def sales_product(request, ids, from_time, to_time, error=None):
...
+        1840,
+        1844,
+        1846,
+        1847,
+        1853,
+        1855,
+        1856,
+        1858,
+        1859,
+    ]
```
hrjakobsen, line 302: Everywhere else maps and lists are compacted but a list of integers are expanded like crazy. hmm..

hrjakobsen (approved):
I don't agree with all of the stylistic choices of black. But I like that the coding conventions are verified in the github actions.

stregreport/models.py
cogitantium: I agree

kiosk/views.py
cogitantium: I do agree

stregreport/views.py
cogitantium: I think it's because it breaks the 120 char limit, since caffeine just above beer does not have this expansion.

cogitantium:
@hrjakobsen I do agree that some formatting choices are less than ideal but I think it's worth it to have a unified code-style.

hrjakobsen:
I assume this will cause difficulties with other open PRs, can we manage that somehow?

cogitantium:
> I assume this will cause difficulties with other open PRs, can we manage that somehow?

The currently open PRs are not too large, so I don't expect the introduction of Black to require more than a single run to make them compliant with the code-style.

.github/workflows/django.yml
```diff
@@ -33,5 +33,5 @@ jobs:
       run: |
         coverage report
     - name: Upload coverage reports to Codecov
-      uses: codecov/codecov-action@v1.0.14
```
VirtualSatai, line 36: Does black only run on ".py" files or can it be setup on other files too?

VirtualSatai (approved):
I think this is great overall. @DelusionalLogic and I used to use `flake8` or `autopep8` to try and keep the file consistent. The narrow (max 80 chars) style was mostly his preference, but anything automatic is fantastic.

VirtualSatai:
How much does this conflict with #179 @cogitantium ?

cogitantium:
> How much does this conflict with #179 @cogitantium ?

It entirely replaces #179, I just closed that PR

.github/workflows/django.yml
cogitantium: It is only a Python formatter and it only runs on non-migration .py files. The options are as follows:
`-t py36 -l 120 --exclude 'migrations' stregsystem stregreport kiosk`

So we aim at py36, allow 120 char line-lengths, and only take the three directories with our three *apps*.

cogitantium: > Does black only run on ".py" files

On second though, that's a great point. Maybe we should have a formatter for html, css, (json/yml)?

.github/workflows/black.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  black:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Black Code Formatter
+      uses: lgeiger/black-action@v1.0.1
+      with:
+        args: " --check -t py36 -l 120 --exclude 'migrations' stregsystem stregreport kiosk"
```
joandrsn, line 18: Since this is a script, I would prefer using the full name of the command line options.
So `-t` becomes `--target-version` and `-l` becomes `--line-length`

joandrsn (commented):
It would be great to have a formatter check on every PR!

I do wonder why not use the command `--skip-string-normalization`. According to [the documentation](https://black.readthedocs.io/en/stable/the_black_code_style.html?highlight=quote#strings), this should only be used for existing projects. I actually kinda like single quotes.

.github/workflows/black.yml
cogitantium: Good idea, fixed in 9c3d9d8

cogitantium:
I have squashed some commits, combined the application commits and put the commit hash in `.git-blame-black-application` such that it can be ignored in future. This information is also contained in CONTRIBUTING.md.

cogitantium:
Since this is a pretty big **commit**ment, I've held out on merging it until all has had an opportunity to discuss. I think three months is enough and will merge it now.

PR #230, opened 2020-11-23
[mobilepaytool] Add permissions to MobilePaytool

cogitantium:
I completely forgot about permissions when writing #204, so I've mirrored the approach that @hrjakobsen made in #197. 

Abstracting batch_payment and mobilepaytool into separate permissions allows stregsystem admins to remove access to - the now legacy - batch payment facility and only allow mobilepaytool access for treo's going forward.


hrjakobsen (approved):
LGTM

PR #232, opened 2020-11-30
Update url to re_path in Django URL patterns

Zinoex:
url is an alias to re_path, and url is likely to be deprecated in the future according to the Django docs.

### DIFF
--- a/kiosk/urls.py
+++ b/kiosk/urls.py
@@ -1,9 +1,9 @@
-from django.conf.urls import url
+from django.urls import re_path
 
 from . import views
 
 urlpatterns = [
-    url(r'^$', views.kiosk),
-    url(r'^random$', views.find_random_media),
-    url(r'^next_real/(?P<item_id>\d+)/$', views.find_next_media_real),
+    re_path(r'^$', views.kiosk),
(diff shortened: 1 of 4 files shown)


joandrsn:
Could you link a source for the deprecation notice in Django docs?

cogitantium:
@Zinoex have you generally tested if urls still redirect as expected? kiosk, debug toolbar, stregsystem-frontend, admin, etc?

I'd like to see the docs as well. I'm all for future-proofing but we're on Django 2.2.13 LTS, so no breaking changes should be expected.

Zinoex:
`url` is literally an alias for `re_path` according to [the Django docs](https://docs.djangoproject.com/en/2.2/ref/urls/#url). The API has been marked for future deprecation since 2.0. I haven't test the redirects yet as I'm not aware how, but as `url` is an alias, it should be a direct drop-in replacement.

Zinoex:
@cogitantium Can we complete this PR?

cogitantium (approved):
@Zinoex, sure, it probably works despite neither of us have tested it. Due to lockdown it'll only be merged onto next for now. Actual release will be whenever we're allowed back at Cassiopeia and I'm able to upgrade.

PR #235, opened 2021-02-28
Add 'total sales monthly' reporting feature for treasurer

falkecarlsen:
During this years VAT bookkeeping it was found that there was no easy way for non-FIT members to look up total sales for any given month. This PR adds that statistic for each of the past 12 months. 

The feature could be extended to allow users to pick any month they like but I chose not to implement that since dealing with time sucks. For the purposes of the treasurer this is more than sufficient for monthly bookkeeping and the yearly bookkeeping associated with paying VAT.

Generates a table viewable only by staff-members:
image

### DIFF
--- a/stregsystem/templates/admin/stregsystem/report/total_sales_monthly.html
+++ b/stregsystem/templates/admin/stregsystem/report/total_sales_monthly.html
@@ -1,0 +1,37 @@
+{% extends "admin/base_site.html" %}
+{% load static %}
+
+{% block title %}Total sales monthly{% endblock %}
+{% block breadcrumbs %}
+    <div class="breadcrumbs"><a href="../..">Hjem</a></div>{% endblock %}
+
+{% block content %}
+    {% load admin_static %}
+
+    <h1>Total sales monthly</h1>
+    <p>Table showing total registered sales during a given month in the past year. Useful for the treasurer for
+        bookkeeping.</p>
+    <table>
+        <thead>
+        <tr>
+            <th>Year</th>
+            <th>Month</th>
+            <th>DKK</th>
+        </tr>
+        </thead>
+        <tr></tr>
+        {% for m in total_sales_monthly %}
+            <tr>
+                <td>
+                    {{ m.0 }}
+                </td>
+                <td>
+                    {{ m.1 }}
+                </td>
+                <td>
+                    {{ m.2 }}
+                </td>
+            </tr>
+        {% endfor %}
+    </table>
+{% endblock %}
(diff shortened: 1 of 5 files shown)


GeneralDenmark (approved):
Linux Great Time Month

stregsystem/views.py
```diff
@@ -6,7 +6,7 @@
 from django.contrib.admin.views.decorators import staff_member_required
 from django.contrib.auth.decorators import permission_required
 from django.conf import settings
-from django.db.models import Q
+from django.db.models import Q, Sum
```
joandrsn, line 9: Shouldnt this change be removed?

joandrsn (approved):
A minor import change. But other than that, looks good!

stregsystem/views.py
cogitantium: Yes, I forgot to remove the import when abstracting query into utils.py. Nice :ok_hand:

cogitantium: Fixed in edd173c

stregsystem/templates/admin/stregsystem/report/total_sales_monthly.html
```diff
@@ -0,0 +1,37 @@
...
+            <th>Year</th>
+            <th>Month</th>
+            <th>DKK</th>
+        </tr>
+        </thead>
+        <tr></tr>
+        {% for m in total_sales_monthly %}
+            <tr>
+                <td>
+                    {{ m.0 }}
```
hrjakobsen, line 26: It would be nicer to have a list of objects instead of a nested list so that you could write

```
{{  entry.year }}
{{ entry.month }}
{{ entry.sales_amount }}
```
or something similar

stregsystem/views.py
```diff
@@ -381,3 +382,19 @@ def qr_payment(request):
...
+def total_sales_monthly(request):
+    data = dict()
+
+    # generate pairs (month, year) from current month, going a year back
+    last_year = datetime.datetime.today().year - 1
+    current_month = datetime.datetime.today().month
+    months = [((m + current_month) % 12 + 1, last_year + (m + current_month) // 12) for m in range(0, 12)]
+
+    # for each month, year pair, sum all sales in that period, and convert numeric month to text
+    data['total_sales_monthly'] = [(y, datetime.date(1970, m, 1).strftime('%B'), p) for y, m, p in
```
hrjakobsen, line 397:
```
import calendar
...
calendar.month_name[m]
```

will get you the month name instead of the strftime hack

hrjakobsen (approved):

stregsystem/templates/admin/stregsystem/report/total_sales_monthly.html
cogitantium: Good idea, fixed in 034eca0

stregsystem/views.py
cogitantium: Good idea, fixed in 034eca0

falkecarlsen:
Inactive

PR #250, opened 2021-09-09
Bugfix remove tox to fix black-action issue

falkecarlsen:
Due to unresolved issues with running black through tox, I suggest removing tox and simply running black locally according to `CONTRIBUTING.md` to adhere to coding standards.


joandrsn (approved):
:shipit:

PR #251, opened 2021-09-12
[mobilepaytool] Add versioning on mobilepayments to counter race-cond…

falkecarlsen:
…itions

Also adds a check for a changed formset in mobilepaytool, to avoid writing form if no changes has been made.


falkecarlsen:
Superceded and incorporated by #253

PR #253, opened 2021-09-17
Bugfix mobilepayment form submit

joandrsn:
This PR should fix the submission of MobilePayment which are already modified in the database.

The added functionality of this PR is that when a `MobilePayment` is ignored, it is only logged on the `MobilePayment` and stregsystemet no longer creates a `Payment` of value 0.
One "feature" of this PR is that when submitting the form, the form automatically removes the rows where status is `UNSET` and where status is `APPROVED` with no `Member` attached. This is to ensure that we only process rows that needs processing and ignore rows with unfinished data entry.


falkecarlsen:
Couldn't figure out how to mock the formset from mobilepaytool in tests but a workaround is setting up `submitted_data` fixtures for tests on the new submit function and emulate mobile payment (autopayment) mutations.

stregsystem/models.py
```diff
@@ -378,29 +388,91 @@ def delete(self, *args, **kwargs):
...
-                f"MobilePayment (transaction_id: {processed_payment.transaction_id})",
+        processed_mobile_payment: MobilePayment  # annotate iterated variable (PEP 526)
+        for processed_mobile_payment in make_processed_mobilepayment_query():
+
+            if processed_mobile_payment.status == MobilePayment.APPROVED:
+                payment = Payment(member=processed_mobile_payment.member, amount=processed_mobile_payment.amount)
+                # Save payment and foreign key to MobilePayment field
+                payment.save()
+                payment.log_from_mobile_payment(processed_mobile_payment, admin_user)
+                processed_mobile_payment.log_mobile_payment(admin_user, "Approved")
```
joandrsn, line 399: Only a minor comment @falkecarlsen but should we really add another logentry for the payment if we approve the `MobilePayment`? The code already creates a `LogEntry` for the payment where it came from.

stregsystem/utils.py
```diff
@@ -83,6 +84,20 @@ def make_unprocessed_member_filled_mobilepayment_query() -> QuerySet:
...
+    # but not other fields if in race condition with autopayment
+    inconsistent_payments = MobilePayment.objects.filter(
+        Q(payment__isnull=False) & Q(status__exact=MobilePayment.UNSET)
+    )
+    for p in inconsistent_payments:
+        p.status = MobilePayment.APPROVED if p.payment.amount > 0 else MobilePayment.IGNORED
+        p.save()
+
+
 def date_to_midnight(date):
```
joandrsn, line 101: Is this function used anywhere? I see it imported into `stregsystem/views.py`, but not used?

joandrsn (commented):
I can't approve my own PR, but I think these minor changes should be addressed.

stregsystem/models.py
joandrsn: Resolved: Keep it as is.

falkecarlsen (approved):
I'm satisfied, think the new submission function covers all cases of race-conditions but would like a second opinion.

We gave up trying to use the Django form-specific clean/validate methods as they did not support raising exceptions and handling them in the mobilepaytool view. This is sorta ugly but works.

Badgie (approved):
LGTM

PR #272, opened 2021-10-04
Add nix-shell setup for easy development

SEbbaDK:
The shell specification uses the current requirements.txt and should add no maintenance.
It makes it very simple for nix users on linux/mac to get ready to work on stregsystemet via `nix-shell`.
We can hopefully remove the two custom packages later when @jbjjbjjbj upstreams the packages to `nixpkgs`


shell.nix
```diff
@@ -0,0 +1,45 @@
...
+	# We split the lines
+	split = builtins.split "\n" reqs;
+	# We remove empty lists and null values
+	lines = builtins.filter (x: builtins.isString x && x != "") split;
+	# Now we have tuples on the form [ name version ]
+	reqtuples = builtins.map (builtins.match "(.*)==(.*)") lines;
+in
+pkgs.mkShell rec {
+    name = "stregsystemet";
+    src = ./.;
```
jbjjbjjbj, line 21: Why add a `src` as this is only a shell.

```suggestion
```

shell.nix
```diff
@@ -0,0 +1,45 @@
+{ pkgs ? import <nixpkgs> {}
+, toLower ? pkgs.lib.toLower
+, python ? pkgs.python39
+, pythonPackages ? pkgs.python39Packages
```
SEbbaDK, line 4:
```suggestion
, python ? pkgs.python36
, pythonPackages ? pkgs.python36Packages
```
Since the README says we should use python 3.6, we should do the same in the shell

django-debug-toolbar.nix
```diff
@@ -0,0 +1,24 @@
+{ pkgs ? import <nixpkgs> {}
+, pythonPackages ? pkgs.python39Packages
```
falkecarlsen, line 2:
```suggestion
, pythonPackages ? pkgs.python36Packages
```

django-select2.nix
```diff
@@ -0,0 +1,26 @@
+{ pkgs ? import <nixpkgs> {}
+, pythonPackages ? pkgs.python39Packages
```
falkecarlsen, line 2:
```suggestion
, pythonPackages ? pkgs.python36Packages
```

falkecarlsen (changes requested):
I've heard it doesn't matter if the 'main' nix-file specifies py-version, but just for good order

Mast3rwaf1z:
@krestenlaust it works with nixos-23.05, but later versions like 23.11 or unstable seems to have checks in place against the older version of django we use.

There seems to be a missing dependency to a module called `django.utils.six`, i don't know what this is.

Also this fork is very outdated

I would suggest that if we want a nix shell, i would prefer it to be written as a flake instead, and rewritten to work with the current system.

krestenlaust:
Thanks you for the insight @Mast3rwaf1z 👌

soupglasses:
Waiting for #409 so we dont need to package ancient versions of Python dependencies. :sunglasses:

Mast3rwaf1z:
@krestenlaust with #459 merged, I think we should close this pr, it basically does the same with less flexibility

krestenlaust:
Fixed by #459

Issue #261, opened 2021-09-29
Framework for theming of frontend


falkecarlsen:
Some FITnisser have previously made themes for the frontend:
![image](https://user-images.githubusercontent.com/11318702/135219367-6b4b6e35-5ba9-4d5b-902f-52838098672b.png)

It would be nice if we could have a framework for defining these instead of making ad-hoc overrides.

I imagine a structured way of defining CSS classes, etc. for elements in the frontend, and then a system for providing different stylesheets to the frontend at render-time.

JakobTopholt:
Leaving a comment to remind myself to consider making an easter egg, where Kurt's C videos sound (if you know you know) plays when you get a stregforbud.

falkecarlsen:
> where Kurt's C videos sound

remember that the stregmachine has never had a speaker

PR #273, opened 2021-10-04
Framework for monthly sprites

MKKlausen:
Basic framework for monthly sprites.


falkecarlsen (changes requested):
It doesn't look like you removed the original `bat.js`, please do that :)

hrjakobsen:
I agree for the need of a generalized framework for easter eggs, so the idea behind this PR is good. I think, however, that the framework should exist on a Django level rather than a frontend level. I've sketched out an idea for this in #261. This approach is still very specialized to sprites flying around the screen, which I can't imagine is something we want for too many months every year. Just think of the poor raspberry pi.

atjn:
Closed in favor of #482

PR #274, opened 2021-10-04
Minimum spec for inventory management

Zaph-x:
This pull request adds a minimal inventory management system to stregsystemet. The idea is to be able to get historical data from stregsystemet about what products sell well, and what products do not sell well.


krestenlaust:
> @krestenlaust consider this a good project for future F-acking events. We have a wireless USB barcode-scanner for use with some frontend. A WIP frontend that is stand-alone is available here https://github.com/joandrsn/inventory

Thanks for the inspiration, having an inventory system would be really awesome! I'll bookmark it for now, but I'll look into it when I get the time.

PR #277, opened 2021-10-06
Ease Of Life

cNorspang:
added 0.0.0.0 to allowed hosts, for easier testing on other devices than host device

### DIFF
--- a/local.cfg.skel
+++ b/local.cfg.skel
@@ -19,5 +19,4 @@ USER =
 PASSWORD =
 
 [hostnames]
-2=127.0.0.1
-3=localhost
+2=*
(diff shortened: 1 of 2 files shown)


VirtualSatai:
> This would add 0.0.0.0 to the allowed hosts on the production server as well, which I don't think we want.
@hrjakobsen 

Does the production server use local.cfg.skel? I would expect that we have previously made a local.cfg that is not tracked in git (https://github.com/f-klubben/stregsystemet/blob/next/.gitignore#L4) that it uses. I don't have access to the server anymore (since AAU closed all port 22 connections) 

The local.cfg.skel is just a skeleton for new users to have a start for their own local.cfg file. 

I could be wrong, but I think changing this file in the repo is fine.

hrjakobsen:
@VirtualSatai previously the PR changed `treo/settings.py` instead of `local.cfg.skel`. The change now should be fine.

PR #278, opened 2021-10-06
Feature making the QR-codes clickable

### LINKED ISSUE #270: Make it so, that QR codes are clickable, for if you're on phone.
If you click the QR code for inserting money, it should open the link on your phone

denn4617:
Closes #270. 
Added a feature which makes the QR-codes clickable. This makes it more convenient for mobile users, since they can't scan the QR. When clicking the QR-code a 'href' will redirect to the mobilepay query, which takes to parameters: username and amount.


falkecarlsen (approved):

falkecarlsen:
> I may be nitpicking, but the PR seems to change formatting which isn't related to the PR. It would be prettier if that was in a separate commit or PR (+1 for Hacktoberfest as well).

You're right but we don't have code-style rules for HTML (yet), so it effectively doesn't matter. If @denn4617 wants to pump their PRs, you are welcome to revert the style changes and do them in another PR.
@denn4617, let me know - then I'll hold off on deploying this :)

hrjakobsen:
> but we don't have code-style rules for HTML (yet)

I would *love* to have an automated style guide, but I don't know of any tool that can handle the Django template language since it essentially allows for arbitrary Django everywhere in the HTML. At least the popular `prettier` doesn't support it https://github.com/prettier/prettier/issues/5581

PR #282, opened 2021-10-08
bumped version on everything

ScumbagDog:
Time to ditch python 3.6~7 and finally update to python3.9.
This includes bumping all dependencies. YOLO.

A part of this is removing all the admin_static tags, as these were deprecated in django 3


krestenlaust:
Closing in favor of #409 which is up-to-date

PR #287, opened 2021-11-24
Feature: run config for jetbrains editor (PyCharm)

AndersSpringborg:
A run config the django server. 

A config for migrating and loading persistent testdata (load testdata fixture)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,6 @@
+[tool.black]
+target-version = ['py36']
+line-length = 120
+exclude ='(migrations|urls\.py)'
+
+skip-string-normalization = true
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[tool.black]
+target-version = ['py36']
+line-length = 120
+exclude ='(migrations|urls\.py)'
+
+skip-string-normalization = true
```
falkecarlsen, line 6: Would be better not to clutter root, if possible put this toml in .run/ or similar

falkecarlsen (changes requested):
Unfortunately the `$PyInterpreterDirectory$` doesn't resolve for me, changing it manually to the default venv makes runners run. I spent some time debugging this and could not find a solution.
`Error running 'migrate': Cannot run program "$PyInterpreterDirectory$/python" (in directory "/home/user/projects/stregsystemet"): error=2, No such file or directory`

This doesn't matter too much for me but I would like other users @Zaph-x to check if it works for them.

CI fails because of black in requirements:
```
 The conflict is caused by:
    The user requested regex==2017.07.28
    black 21.11b1 depends on regex>=2021.4.4
```

PR #291, opened 2021-12-03
Fix conditional for snowflakes to match one pr day in December

falkecarlsen:


stregsystem/static/stregsystem/snow.js
```diff
@@ -1,7 +1,7 @@
 d = new Date();
 
 if(d.getMonth() === 11){
-    for(let snowflakes=0; snowflakes <= Math.min(d.getDate(), 24); snowflakes++){
+    for(let snowflakes=0; snowflakes < Math.min(d.getDate(), 24); snowflakes++){
```
SEbbaDK, line 4: I think we should add concurrency to this to speed up processing of snowflakes.

SEbbaDK (approved):
Nice B)

stregsystem/static/stregsystem/snow.js
falkecarlsen: Great, write an issue for it B)

Rawrior: I agree, this is an essential improvement.

PR #295, opened 2022-02-11
Add named IDs to stregsystemet

Zaph-x:
This pull request adds named IDs to stregsystemet.
This has been requested by several fembers since STS got named IDs as shorthands for buying.
The named IDs do not change the way products are parsed when purchases, as they are handled in pre-processing. This means the user does not need to know about the IDs, however may use them if they are known to the user.
An example of a named ID purchase is `tester øl` which would be preprocessed to `tester 14` by looking for a named ID in the database linked to øl. If an ID is not found, the string is left as is (e.g. `tester asdfghjkl` would remain the same) and give a parse error.


pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[tool.black]
+line-length = 120
+skip-string-normalization = 1
+exclude = '/(migrations|urls\.py|venv|.venv)/'
+include = '\.pyi?$'
+target-version = ['py37']
```
hrjakobsen, line 6: I feel like this should be a separate PR. Isn't related to named IDs

stregsystem/models.py
```diff
@@ -592,6 +592,17 @@ def is_active(self):
...
 
+class NamedProduct(models.Model):
+    name = models.CharField(max_length=50)
+    product = models.ForeignKey(Product, on_delete=models.CASCADE, related_name='named_id')
+
+    def __str__(self):
+        return self.name + " -> " + self.product.name
+
+    def save(self, *args, **kwargs):
+        super(NamedProduct, self).save(*args, **kwargs)
```
hrjakobsen, line 603: Isn't this the default behaviour? Why do we need to override `save`?

stregsystem/urls.py
```diff
@@ -25,9 +25,14 @@
     re_path(r'^(?P<room_id>\d+)/$', views.index, name="menu_index"),
     re_path(r'^(?P<room_id>\d+)/sale/$', views.sale, name="quickbuy"),
     re_path(r'^(?P<room_id>\d+)/sale/(?P<member_id>\d+)/$', views.menu_sale, name="menu"),
-    re_path(r'^(?P<room_id>\d+)/sale/\d+/\d+/$', lambda request, room_id: redirect('menu_index', room_id=room_id), name="menu_sale"),
+    re_path(
+        r'^(?P<room_id>\d+)/sale/\d+/\d+/$',
+        lambda request, room_id: redirect('menu_index', room_id=room_id),
+        name="menu_sale",
+    ),
```
hrjakobsen, line 32: Shouldn't be unfolded by `black`

stregsystem/views.py
```diff
@@ -74,6 +75,26 @@ def index(request, room_id):
     return render(request, 'stregsystem/index.html', locals())
 
 
+def dump_named_items(request):
+    items = NamedProduct.objects.all()
+    items_dict = {item.name: item.product.id for item in items}
+    return JsonResponse(items_dict, json_dumps_params={'ensure_ascii': False})
```
hrjakobsen, line 81: This looks like a debug page that we shouldn't include

stregsystem/views.py
```diff
@@ -74,6 +75,26 @@ def index(request, room_id):
...
+
+
+def _pre_process(buy_string):
+    items = buy_string.split(' ')
+    _items = [items[0]]
+
+    for item in items[1:]:
+        if type(item) is not int:
+            _item = NamedProduct.objects.filter(name=item.split(':')[0].lower() if ':' in item else item)
```
hrjakobsen, line 90: Adding a `NamedProduct` with a name including ":" or " " would break this function, we should probably check for that somewhere.

treo/settings.py
```diff
@@ -218,29 +216,17 @@
...
-    'loggers': {
-        '': {
-            'level': cfg.get('logging', 'LEVEL'),
-            'handlers': json.loads(cfg.get('logging', 'HANDLERS'))
-        }
-    }
-}
\ No newline at end of file
+    'loggers': {'': {'level': cfg.get('logging', 'LEVEL'), 'handlers': json.loads(cfg.get('logging', 'HANDLERS'))}},
+}
```
hrjakobsen, line 232: Not relevant to this PR

treo/urls.py
```diff
@@ -24,14 +24,14 @@
...
 ]
 
 urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
 
 if settings.DEBUG:
     import debug_toolbar
+
     urlpatterns = [
         re_path(r'^__debug__/', include(debug_toolbar.urls)),
     ] + urlpatterns
```
hrjakobsen, line 37: Not relevant to this PR

hrjakobsen (commented):
I haven't tested this, only read the code changes, and left some comments based on that. I think some more validation of the names of `NamedProduct` are required.

VirtualSatai (changes requested):

stregsystem/models.py
```diff
@@ -592,6 +593,20 @@ def is_active(self):
...
+    product = models.ForeignKey(Product, on_delete=models.CASCADE, related_name='named_id')
+
+    def __str__(self):
+        return self.name + " -> " + self.product.name
+
+    def save(self, *args, **kwargs):
+        self.name = self.name.strip()
+        if re.match(r'(^\d+$|.*?\s|.*?:)', self.name):
+            raise RuntimeError("Name can not include spaces, ':' or be a number")
+        super(NamedProduct, self).save(*args, **kwargs)
```
hrjakobsen, line 607: This should be implemented as a [validator](https://docs.djangoproject.com/en/3.2/ref/validators/) for the `name` field

hrjakobsen (changes requested):

joandrsn (approved):
:shipit:

hrjakobsen (approved):
I think alias handling in views could be handled better. 

* Why do we show a random alias instead of all? 
* Why do we *have* to show an alias if none exist? (N/A -> id)

However this is nothing that blocks merging.

Issue #254, opened 2021-09-17
Make tilmeldings-flow for stregsystemet


falkecarlsen:
It would be nice if TREO's weren't required to manually process each new member, as they have to take into account both MobilePay backend for amount, already-taken usernames, and any missing information.

Since introducing #204 this issue becomes somewhat easier, since we can now automatically get and process optankninger, we could automagically process tilmeldinger.

One (probably not the best) approach could be:
- Make new page with modelform on Member (make check for already-taken username). Importantly still require last `n` characters of MobilePay transaction for tilmelding.
- Make cronjob which processes these tilmeldinger but only when a MobilePayment has been received with matching last `n` transaction id and sufficient amount of money.

jonasKjellerup:
Currently working on an implementation for this. My current plan for the sign-up flow is as follows:

 1) The new user fills out a form with their user information (username, first/last name and email).
    The username and email are verified to be unique. If the information is ok we create a record
	with a unique token.
 2) We present the user with a QR-code/link for a mobilepay transaction with a per-composed description
    and amount. The description contains the aforementioned token, which is used to identify the correct transaction.
    (When doing it this way we can avoid having the user copy the last digits of the transaction id)
 3) Once the full amount has been payed, the mobilepay transaction will at some be imported at which point
    it will be scanned for the token and the user account activated.
	
As for handling the mobilepay transactions are there any preferences/guidelines for how we handle these
kind of things? E.g. an adittional cron job Vs. hooking into the existing job through an event bus or similar.

falkecarlsen:
@jonasKjellerup sounds interesting - go for it!
Just expose a Django command, similar to how autopayment and import is implemented currently and we'll stage it in production as cronjobs. Keeping it modular also helps avoid these payment-components from breaking eachother. While having all MobilePay integration event-driven would be nice, we only pull transactions on a cronjob due to API limitations, so it really doesn't matter that subsequent modules are event-driven atm.

jonasKjellerup:
I have a working, albeit slightly naively implemented, prototype at this point. It still needs further testing but it does seem to work as intended thus far.

> Consider modifying the Member model to contain this new data,

I ended up storing the new data in a seperate model/table (consisting of a Member reference, a token and the amount that is due to be payed). I didn't see much point in expanding the Member model since the additional information isn't relevant for existing members.

> keep in mind that the 'active' field does not entirely fit your use-case as it's for banning/no-spam etc. - not for whether the membership is paid up yet.

[...]

PR #297, opened 2022-03-11
Feature signup

jonasKjellerup:
The pull request addresses issue #254 

The pull request adds the following:

- A signup form (linked to via "Tilmeld" button on the index page)
- Adds a `Command` for automatically processing signup payments

The new signup flow would be as follows:

1. The user fills out the signup form.
2. On success the user will be redirected to a status page with a clickable (for mobile users) QR-code.
3. Once payment has been registered the account will be activated and a welcome mail will be sent.


stregsystem/views.py
```diff
@@ -465,3 +472,43 @@ def qr_payment(request):
     data = 'mobilepay://send?{}'.format(urllib.parse.urlencode(query))
 
     return qr_code(data)
+
+
+def signup(request):
+    is_post = request.method == "POST"
+    form = SignupForm(request.POST) if is_post else SignupForm()
```
falkecarlsen, line 479: Redundant, can be reduced to one statement. If request.POST, then it is post.

stregsystem/forms.py
```diff
@@ -35,6 +35,19 @@ class PurchaseForm(forms.Form):
     product_id = forms.IntegerField()
 
 
+class SignupForm(forms.Form):
+    username = forms.CharField(max_length=16, label='Brugernavn',
+                               error_messages={'required': 'Udfyldning af `Brugernavn` er påkrævet.',
+                                               'max_length': 'Længden af `Brugernavn må ikke overstige 16 tegn.`'})
```
Nielswps, line 41: Misplacement of '\`', should be 'Længden af `Brugernavn` må ikke overstige 16 tegn.'

stregsystem/forms.py
```diff
@@ -35,6 +35,19 @@ class PurchaseForm(forms.Form):
...
+    email = forms.EmailField(label='E-Mail')
+    firstname = forms.CharField(max_length=20, label='Fornavn',
+                                error_messages={'required': 'Udfyldning af `Fornavn` er påkrævet.',
+                                                'max_length': 'Længden af `Fornavn må ikke overstige 16 tegn.`'})
```
Nielswps, line 45: Same as above, should be 'Længden af `Fornavn` må ikke overstige 16 tegn.'

stregsystem/forms.py
```diff
@@ -35,6 +35,19 @@ class PurchaseForm(forms.Form):
...
+    lastname = forms.CharField(max_length=30, label='Efternavn',
+                               error_messages={'required': 'Udfyldning af `Efternavn` er påkrævet.',
+                                               'max_length': 'Længden af `Efternavn må ikke overstige 16 tegn.`'})
```
Nielswps, line 48: Same as above, should be 'Længden af `Efternavn` må ikke overstige 16 tegn.'

stregsystem/migrations/0016_auto_20220131_2109.py
```diff
@@ -0,0 +1,29 @@
...
+        migrations.AddField(
+            model_name='member',
+            name='signup_due_paid',
+            field=models.BooleanField(default=True),
+        ),
+        migrations.CreateModel(
+            name='PendingSignup',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('due', models.IntegerField(default=20000)),
```
Nielswps, line 24: Consider adding a comment about what 20000 represents. I assume that it is 200,00 DKK in øre?

stregsystem/models.py
```diff
@@ -661,3 +666,30 @@ def __unicode__(self):
 
     def __str__(self):
         return self.title + " -- " + str(self.pub_date)
+
+
+class PendingSignup(models.Model):
+    member = models.ForeignKey(Member, on_delete=models.CASCADE, null=False)
+    due = models.IntegerField(default=200 * 100)
```
Nielswps, line 673: IMO: Either write 20000 as in `stregsystemet/models.py` or do this there as well, to be consistent

Nielswps (changes requested):
I think this would be a brilliant feature during events for members where non-members might attend, wanting to signup, good job :+1: 
Just some minor string typos

stregsystem/models.py
jonasKjellerup: I assume you mean as in `[stregsystem/migrations/0016_auto_20220131_2109.py](https://github.com/f-klubben/stregsystemet/pull/297/files/90af2f04a8eb117a6597404182d6616b5036601b#diff-ff4bbb2f164add9b3111918d75b6a5830763528bfe98589659b83c3c58099373)`. The migrations files are auto generated, hence the discrepancy based on the class from the model file, so it doesn't really make sense to edit them.

krestenlaust:
Fixes #254

stregsystem/models.py
```diff
@@ -673,3 +678,30 @@ def __unicode__(self):
...
```
krestenlaust, line 685: Wouldn't it be better to define the membership cost outside of the model to not require migrations in-case of vedtægtsændringer?

krestenlaust:
I've just looked into this implementation to make sure that #407 doesn't make this outdated, it doesn't look like it. It just needs to catch up with upstream (66 commits behind)

JakobTopholt:
This pull request would be a godsent.

stregsystem/models.py
krestenlaust: I guess it's just a default

krestenlaust:
Working towards the following feature:
At a meeting with the forstyrrelse, it was decided that it would be preferable to have a view like MobilePayTool to accept signups manually

krestenlaust (commented):
Initial reviews are welcome🙌

Major changes in this PR:
 - Introduced new model PendingSignup, which only exists to track members who isn't approved and hasn't paid membership due.
 - Turned existing Mobilepaytool abstract, and implemented a signup tool with the same concept.
 - Some refactoring of the AutoPayment-command to properly process paying towards membership dues.

Missing features:
 - [ ] 'Area of Study'-choice in the signup form to populate 'notes' of the member.
 - [ ] ~~'Reject'-functionality for Signup-Tool hasn't been implemented yet.~~ **Feature rejected for now.**
 - [ ] ~~Haven't added QR-code to error_signupdue-screen (Presented when trying to access user menu with member without due paid)~~ **This feature won't be implemented here.**

Known bugs:
 - [x] ~~Approving a member who has already paid their due, will not remove the pending signup-object.~~ **Fixed**
 - [ ] I also suspect there are some issues related to email triggers, depending on the order a user is approved and due is paid.

I'll try to add some more email tests

krestenlaust:
Feature is finished! I've decided to ignore the potential bug with welcome emails for now.

BoAnd:
As I see it the sign up logic works. There is some small problem with the differences sign up templates.
Like we could merge template **error_signupdue.html** and **error_signup_not_approved.html** into one template. 

But I think it could be nice to just merge this into prod and the fix the problems afterwards (If there is any :dagger:).

BoAnd (approved):

krestenlaust:
> As I see it the sign up logic works. There is some small problem with the differences sign up templates. Like we could merge template **error_signupdue.html** and **error_signup_not_approved.html** into one template.
> 
> But I think it could be nice to just merge this into prod and the fix the problems afterwards (If there is any 🗡️).

Thank you for your review! I've made another issue with your note

PR #299, opened 2022-03-31
Easter theme

miniJmo:
yeah, the title more or less says it all.
I added an easter theme that is more or less a copy-paste of the Christmas theme but with a "cursed" easter egg that is activated once a day for a minute at 13:37 if the page is reloaded in that timeframe.


falkecarlsen (approved):
LGTM :shipit:

PR #302, opened 2022-04-12
Feature tickets

JohanThomsen:
A feature which allows someone to attach an event to product, which will the be shown in the "Mine Billeter" tab found on the userprofile, allowing a user to easily keep track of and remember exact times of events they have bought tickets for.

The page can be found by going to the user profile and clicking "Mine Billeter"
minebilleter
In here You can see the tickets the person has bought:
Billetside

The tickets are created by attaching an event to a product, using the following form:
tomAddEvent
Name = The name of the event (F-acking for example)
Active = Whether the event is active or nor
Time = The start of the even
Ticket start Threshold = The time from which the ticket will be visible on the "mine billeter" page
Ticket end Threshold = The time at which the ticket will no longer be visible on the "mine billeter" page

Here is how it could look when filled out
FuldAddEvent

Tests are also added


stregsystem/views.py
```diff
@@ -218,6 +219,29 @@ def menu_userinfo(request, room_id, member_id):
     return render(request, 'stregsystem/menu_userinfo.html', locals())
 
 
+# Used to see active tickets
+def menu_ticketsview(request, room_id, member_id):
+    room = Room.objects.get(pk=room_id)
+    member = Member.objects.get(pk=member_id, active=True)
+    event_list = Event.objects.all()
+    kek = 2
```
falkecarlsen, line 227:
```suggestion
```

stregsystem/templates/stregsystem/menu_ticketsview.html
```diff
@@ -0,0 +1,35 @@
...
+    <td>{{event.name}}</td>
+    <td>{{event.product.name}}</td>
+    <td>{{event.product.price|money}}</td>
+    <td>{{event.time}}</td>
+</tr>
+{% endfor %}
+{% endautoescape %}
+</table>
+</center>
+<h1>{{kek}}</h1>
```
falkecarlsen, line 33: eksdee 
image

falkecarlsen (changes requested):
Looks good! Seems like tests are mostly just yoinked from previous Product tests - that's probably fine.

hrjakobsen (changes requested):

falkecarlsen:
@JohanThomsen will you resolve the comments from @hrjakobsen, then we'll deploy it soon:tm:

PR #313, opened 2022-06-01
Added notice below custom mobilepayment QR code

ScumbagDog:
Added notice that 50dkk is the minimum allowed payment on the payment page, so users don't try to pay less.

Yes, UX is my passion 😎


falkecarlsen (approved):
Ingen ugler i mosen

Issue #303, opened 2022-04-21
Make QR code always on top


Zaph-x:
Currently during April, the QR codes become un-scannable, due to overlapping imagery.
This prevents scanning of QR codes, leading to manual MP transactions, which do not contain the username of a user. A z-index attribute along with absolute positioning, could resolve this issue
![image](https://user-images.githubusercontent.com/35431127/164396811-86d450aa-bef0-484a-9760-94ffbf75e214.png)

jonasKjellerup:
Fixed in #315

PR #315, opened 2022-10-05
Prevent bats/flakes from overlapping QR-codes

jonasKjellerup:
Addresses issues #303.

- Sets relative positioning and z-index of QR-codes on the user/pay page.
- Tested for bats, beerflakes, and snowflakes.


VirtualSatai (approved):
Would be more fun to always have a bat in the QR code, but it not fucking it up because of the redundancy allowed in QR codes. Anyway, this LGTM :shipit:

Zaph-x (approved):

falkecarlsen (approved):
:shipit:

PR #316, opened 2022-10-12
Upgrade HTML doctype in stregsystem

atjn:
Browser engineers have spent the last 10 years writing speed optimizations for HTML. Those optimizations are all disabled for the stregsystem because it still uses the old XHTML 1.1 doctype.

Updating to HTML 5 could make the interface run noticeably faster on the cpu-constrained raspi/pos-thing that hosts the stregsystem interface at strandvejen.

I have tested this change and everything seems to work. I also don't see why anything should break from it.

### DIFF
--- a/stregsystem/templates/500.html
+++ b/stregsystem/templates/500.html
@@ -1,15 +1,10 @@
 {% load static %}
-<?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN"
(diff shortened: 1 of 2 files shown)


falkecarlsen (approved):
Looks fine, not sure if it will improve performance but let's try :shipit:

atjn:
Cool! Even if it doesn't improve performance, it is still good practice, and fixes a bunch of annoying bugs from the older HTML versions ¯\\_(ツ)_/¯

PR #317, opened 2022-10-12
Include chosen date in rank

BaBrixx:
When accessing the user rankings, the statistics does not include the date specified in the "To" date field. 

This PR makes sure that the "to_date" includes the specified date.


falkecarlsen (approved):
Works on my machine :shipit: . There's something funky with `category_per_uni_day` for which it states 0.07 units/day for one month but also 2.25 units/day for a single day, with one purchase in the timeframe. If interested, you can look into that.

PR #318, opened 2022-10-13
Python3 for Ubuntu requires `venv` separately

atjn:
Tiny fix in docs - newer versions of `python3` do not bundle `python3-venv`, it needs to be manually installed.


falkecarlsen (approved):
:shipit:

PR #319, opened 2022-10-17
Feature TOS

### LINKED ISSUE #301: Make TOS notice
Something that let's us be slightly more above-board.

When #297 or equivalent is implemented, this needs to be done for signup as well. 
At some point, when contact-less payment by way of study-card is implemented, we need a notice to be allowed to store their six-digit study-card-number.

falkecarlsen:
Resolves #301.

Should be expanded/checked to strictly adhere to GDPR. 
I'm unsure as to the phrase "By continuing to use x you agree to y" is strictly legal but we are in a predicament as our current 4463 members have not agreed to this TOS, so we need an onboarding procedure with a active choice (the active choice being the legal requirement afair).
In future, when #297 is finalised and merged, we can do a simple required-true checkbox at signup for this TOS, to ensure acceptance for all further members.


atjn:
I have written a few corporate GDPR compliance reports and can confirm that "By continuing to use x you agree to y" is legally valid in the same way that HTML is a programming language.

falkecarlsen:
Then we might want to do an on-boarding process for existing members, e.g. a pop-up when non-TOS member tries to do a purchase. Spamming all members with accept-requests might not be what we want to do.

atjn:
> Then we might want to do an on-boarding process for existing members, e.g. a pop-up when non-TOS member tries to do a purchase.

That is the only correct way to do it. Although you can discuss if it is really necessary for this situation.

You also have to consider that if you want to update the TOS, you need a way to inform all existing users that it has been updated. An update does not require active consent from users, as long as you are reasonably sure that they saw the message about the updated TOS.
You could develop the notification feature now, and use that to inform existing users that you are updating the TOS from version `-1.0 lol` to version `1.0`. It isn't as good legally as the active consent that you mention above, but I doubt anyone would be mad about this approach.

Nielswps:
Is it too much to add a column in the `members` table for `toc_accepted` (possibly with a timestamp, naming it `toc_accepted_at` instead)? Then all users that have not actively accepted the TOC could be prompted before any actions are completed (the timestamp would allow us to check whether they have seen/passively accepted the updated TOC).

krestenlaust (approved):
LGTM IANAL

atjn:
Closing in favor of #525

PR #320, opened 2022-10-19
Added api endpoints

Zaph-x:
This pull request aims to get us closer to moving away from a monolithic application.
This is done by creating API endpoints, which exposes mostly public facing data in stregsystemet through GET requests.
Furthermore, a sale endpoint which accepts POST requests has been added.

### DIFF
--- a/stregsystem/views.py
+++ b/stregsystem/views.py
@@ -574,21 +574,20 @@ def dump_named_items(request):
 
 @csrf_exempt
 def api_sale(request):
-    if request.method == 'POST':
+    if request.method != "POST":
+        return HttpResponseBadRequest()
+    else:
         data = json.loads(request.body)
-        buy_string = data['buystring'].strip()
-        room = data['room'] or None
-        member_id = data['member_id'] or None
+        buy_string = str(data['buystring']).strip()
+        room = str(data['room']) or None
+        member_id = str(data['member_id']) or None
 
-        if not type(room) is int:
-            return HttpResponseBadRequest("Invalid room")
-
+        if room is None or not room.isdigit():
+            return HttpResponseBadRequest("Missing or invalid room")
         if buy_string is None:
             return HttpResponseBadRequest("Missing buystring")
-        if member_id is None:
-            return HttpResponseBadRequest("Missing member_id")
-        elif not member_id.isdigit():
-            return HttpResponseBadRequest("Invalid member_id")
+        if member_id is None or not member_id.isdigit():
+            return HttpResponseBadRequest("Missing or invalid member_id")
 
         try:
             username, bought_ids = parser.parse(_pre_process(buy_string))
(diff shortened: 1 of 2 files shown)


stregsystem/views.py
```diff
@@ -487,3 +483,174 @@ def qr_payment(request):
...
+    member = Member.objects.get(pk=member_id, active=True)
+    return JsonResponse({'active': member.active})
+
+
+def convert_username_to_id(request):
+    username = request.GET.get('username') or None
+    if username is None:
+        return HttpResponseBadRequest("Missing username")
+    member = Member.objects.get(username=username)
+    return JsonResponse({'member_id': member.id})
```
falkecarlsen, line 516: No check for whether DRM returns object, also for all other endpoints

stregsystem/views.py
```diff
@@ -487,3 +483,174 @@ def qr_payment(request):
...
+    return JsonResponse({'balance': member.balance, 'username': member.username, 'active': member.active, 'name': f'{member.firstname} {member.lastname}'})
+
+
+def dump_named_items(request):
+    items = NamedProduct.objects.all()
+    items_dict = {item.name: item.product.id for item in items}
+    return JsonResponse(items_dict, json_dumps_params={'ensure_ascii': False})
+
+
+@csrf_exempt
```
falkecarlsen, line 560: Would be nice if we *could* retain this

stregsystem/views.py
```diff
@@ -487,3 +483,174 @@ def qr_payment(request):
...
+        try:
+            member = Member.objects.get(pk=member_id, active=True)
+        except Member.DoesNotExist:
+            return HttpResponseBadRequest("Invalid member_id")
+
+        if username != member.username:
+            return HttpResponseBadRequest("Username does not match member_id")
+
+        if len(bought_ids):
+            room = Room.objects.get(pk=room)
```
falkecarlsen, line 591: Needs error check if room does not exist

falkecarlsen (changes requested):
Needs some polish and **definitely** tests but cool :+1:

stregsystem/views.py
Zaph-x: As discussed in person, it would be difficult to retain it at this state. If we were to retain it, we would need to store cookies, which is what this PR tries to avoid. This is because I wish to avoid making as many requests with applications such as STS.

stregsystem/views.py
```diff
@@ -487,3 +476,249 @@ def qr_payment(request):
...
+
+def dump_named_items(request):
+    items = NamedProduct.objects.all()
+    items_dict = {item.name: item.product.id for item in items}
+    return JsonResponse(items_dict, json_dumps_params={'ensure_ascii': False})
+
+
+@csrf_exempt
+def api_sale(request):
+    if request.method == 'POST':
```
jonasKjellerup, line 577: This should return a `HttpResponse` of some sort, since a GET request currently results in a internal server error. Should ideally result in a 405 Method Not Allowed.

stregsystem/views.py
```diff
@@ -487,3 +476,249 @@ def qr_payment(request):
...
+
+        if not type(room) is int:
+            return HttpResponseBadRequest("Invalid room")
+
+        if buy_string is None:
+            return HttpResponseBadRequest("Missing buystring")
+        if member_id is None:
+            return HttpResponseBadRequest("Missing member_id")
+        elif not member_id.isdigit():
+            return HttpResponseBadRequest("Invalid member_id")
```
jonasKjellerup, line 591: The typing of these parameters is a bit inconsistent. Both `room` and `member_id` are supposed to be numbers, but `member_id` is required to be passed as a string (since `.isdigit()` does not exist for integers - this also causes a internal server error) and `room` is required to be passed as a number.

Ideally it should just parse them if necessary and accept the as if when given integers. But at a minimum it should be consistent.

stregsystem/views.py
```diff
@@ -487,3 +476,249 @@ def qr_payment(request):
...
+            username, bought_ids = parser.parse(_pre_process(buy_string))
+        except parser.ParseError as e:
+            return HttpResponseBadRequest("Parse error: {}".format(e))
+        try:
+            member = Member.objects.get(pk=member_id, active=True)
+        except Member.DoesNotExist:
+            return HttpResponseBadRequest("Invalid member_id")
+
+        if username != member.username:
+            return HttpResponseBadRequest("Username does not match member_id")
```
jonasKjellerup, line 603: Is there a reason for requiring a user identifier both as a parameter in the post body and as a part of the buystring?

jonasKjellerup (changes requested):
The handling of error states is in some cases a bit inconsistent (there are quite a few conditions that cause internal server error response).

Otherwise most things seem to work as expected (based on some manual testing)

jonasKjellerup (approved):

PR #327, opened 2023-03-23
Feature forbrug

falkecarlsen:


falkecarlsen:
This feature has previously been discussed in #177 and elsewhere. My impression is that there's a majority who would like to have this information available.

I'm submitting it now because I just passed 10,000 DKK spent! 🎉

jonasKjellerup (approved):
Looks good

PR #340, opened 2023-08-16
Refactored mail.py reduce code duplication

### LINKED ISSUE #187: Remove hardcoded mail server & mail adress
Came from PR #185.
Currently there are lines of codes like these:

https://github.com/f-klubben/stregsystemet/blob/babdef240deee76ec294265d20fdebd618e2da55/stregsystem/utils.py#L113-L114
and
https://github.com/f-klubben/stregsystemet/blob/babdef240deee76ec294265d20fdebd618e2da55/stregsystem/utils.py#L83

In order to make the application more flexible, there could be a config file with the information.

krestenlaust:
I've reduced code duplication in mail.py. ~~Draft for now, as I'll need to see if I can improve it further~~ (Should also fix #187 )


stregsystem/mail.py
```diff
@@ -12,43 +12,37 @@
...
-
-    context = dict()
-    context.update(vars(member))
-    context.update({'formatted_amount': money(amount)})
-    # TODO: not sure if escape serves any purpose here, since it's already being passed through a template.
-    context.update({'mobilepay_comment': escape(mobilepay_comment)})
-
-    target_template = "deposit_manual.html" if mobilepay_comment else "deposit_automatic.html"
+    msg['Subject'] = subject
     html = render_to_string(f"mail/{target_template}", context)
```
falkecarlsen, line 38: Not specifically related to your refactoring but it might be problematic that we don't statically check the existence of our templates. We only have three at the moment, so it's trivial.

I imagine a try-catch on loading template from disk. On error, staff should be notified in some way (mail as logs are read at most twice a decade). 

Make an issue/PR if you feel this is interesting.

falkecarlsen (approved):
LGTM :shipit:

PR #342, opened 2023-08-18
Added start- and enddates for kiosk items

### LINKED ISSUE #288: Kiosk: Add start+end date for item
We charge companies for ads on Kiosk for a given period. Currently, we rely on volunteers who have good memory to know when to decommission items - it would be nice if Kiosk items functioned similarly to Products with active-bool being set by start- and end-date.

krestenlaust:
Resolves #288 

- ~~I'm not sure how to format django filters properly~~ ran black
- Change requires migrations to database (I think), unsure about whether `start_datetime` should be nullable as well

I specifically chose datetime instead of date because time of day might be relevant. Maybe change `uploaded_date` to datetime as well.


kiosk/models.py
```diff
@@ -15,8 +16,13 @@ class KioskItem(models.Model):
...
+    start_datetime = models.DateTimeField(default=timezone.now(), blank=True)
+    end_datetime = models.DateTimeField(null=True, blank=True)
 
     @property
     def is_image(self):
         name, extension = os.path.splitext(self.media.name)
         return extension in valid_images
+
+    def is_active_period(self, current_time) -> bool:
+        return (self.start_datetime < current_time) and (self.end_datetime is None or self.end_datetime > current_time)
```
falkecarlsen, line 28:
```suggestion
    def is_active_period(self) -> bool:
        now = timezone.now()
        return (self.start_datetime <= now) and (self.end_datetime is None or self.end_datetime >= now)
```
No need to pass time to filter as arg. May be `@property` as well

kiosk/views.py
```diff
@@ -15,7 +17,12 @@ def find_random_media(request):
     """
     Randomly get a media and return the relative url
     """
-    item = KioskItem.objects.filter(active=True).order_by('?').first()
+    item = (
+        KioskItem.objects.filter(active=True, start_datetime__lte=timezone.now())
+        .filter(Q(end_datetime__isnull=True) | Q(end_datetime__gt=timezone.now()))
```
falkecarlsen, line 22: Django doesn't support passing properties to the ORM filter. You could consider using the filter made in kiosk/models `is_active_period()` to filter all KioskItems:
```py
active_items = random.shuffle([x for x in KioskItem.objects if x.is_active_period()])
if active_items.len > 0:
    return active_items[0]
else:
    raise Http404("No active kiosk items found") 
```

But as I'm writing this suggestion, I realise this isn't better. In that case, consider removing the unused `is_active_period()`

falkecarlsen (changes requested):
Please make sure that timezones are taken into account with these. AFAIR, we run *and* present UTC.
If it turns out that we are using only UTC already, continue that.

kiosk/models.py
krestenlaust: My thought process was that it could be used to introduce a mockup time in case of a test, but there's probably not much use for that.

falkecarlsen: Look at our usage of `freezegun` for mocking time in tests: https://github.com/f-klubben/stregsystemet/blob/9748ebebecc7cdb8fca69457aad70fa1a86924a2/stregsystem/tests.py#L823

Not needed for this feature as the queries are so simple.

falkecarlsen:
Kiosk entries are typically timeframe-bound information, either internal events or sponsorships from industry which we sell for a period of time. Previously, we've forgot to stop sponsorships because nobody knew whether we were still getting paid for it and similarly, internal events has been advertised long after the event has stopped.

We might want to explicitly disallow not setting an end-time, such that this issue is mitigated. Add a tooltip for the admin interface to the tune of "Set end time to day after event".

I suggest requiring end-time, and allow blank start-time, but setting timezone.now() as default if unset.

VirtualSatai:
Cool addition to the kiosk I didn't think of when I implemented it 👍 

I dislike putting too many restrictions on the usage of the kiosk, I think it's better to trust the TREO / forstyrelse to admin it. I think an unbounded endtime is fine, otherwise you'd just have to put in a stupidly late time for the generic items (like the "sign up for the f-klub"-ad). Defaulting start-time to timezone.now() is a fine choice if you really want to disallow null.

krestenlaust:
I agree. As far as I understand, it's not uncommon to have indefinite kiosk entries. There could also be a case where an end date is unknown at upload time, having a nullable entry also communicates a missing end date in case of null.

The current PR implementation makes both date fields nullable. The advantage I see, is it leaves all existing entries untouched. I believe the semantic value of nullable fields is worth more, than avoiding null for the sake of avoiding it.

Furthermore, the reason I don't find a default value compelling, is that the field 'created at' already exists.

Though, I haven't experienced the kiosk in action, and I haven't used it as an admin either. (Only locally)

falkecarlsen:
I'm fine with either approach, my argument is to force thoughtfulness when picking end_date. Leaving it blank is a much easier error to make, than setting date to >=anno 2040.

My hope is that this could mitigate entries running for longer than they're meant to. I don't subscribe to the:
> There could also be a case where an end date is unknown at upload time

comment as an argument for allowing nullable end_date as it allows admins to make _temporary_ settings for kiosk entries - and as we especially know in stregsystemet - there is nothing as permanent as a temporary solution.

The above comments are nitpicky, we should just decide on one of the two approaches :shipit:

falkecarlsen (approved):
:shipit:

PR #347, opened 2023-08-25
added a random function to randomly distribute news entries

### LINKED ISSUE #346: Make news to get random entry like kiosk
https://github.com/f-klubben/stregsystemet/blob/9748ebebecc7cdb8fca69457aad70fa1a86924a2/stregsystem/views.py#L55

Multiple news might want to be shown at same time, randomly pick upon load which one to show. 

Currently, if more than one exist for some timestamp, we get an error on ORM filter `.get()` which blocks stregsystemet from loading.

Dev-Destructor:
fixes #346


VirtualSatai (approved):
LGTM :shipit:

stregsystem/views.py
```diff
@@ -54,7 +55,10 @@
 
 def __get_news():
     try:
-        return News.objects.filter(stop_date__gte=timezone.now(), pub_date__lte=timezone.now()).get()
+        current_time = timezone.now()
+        news_items = News.objects.filter(stop_date__gte=current_time, pub_date__lte=current_time)
```
falkecarlsen, line 59:
```suggestion
        news_items = News.objects.filter(stop_date__gte=current_time, pub_date__lte=current_time).order_by('?').first()
```
https://docs.djangoproject.com/en/dev/ref/models/querysets/#django.db.models.query.QuerySet.order_by

falkecarlsen (changes requested):
Doesn't pass tests as the empty set, as random.choice([]) is an error

Dev-Destructor:
> Doesn't pass tests as the empty set, as random.choice([]) is an error

Yeah, Sorry

falkecarlsen (approved):
Removed unnecessary random.choice call on list of one elem.

Thanks :shipit:

PR #349, opened 2023-08-31
Redumped fixture, and added a news entity

krestenlaust:
This PR updates the common mobilepay-fixture to include a news entity.

Issue origin:

    You can load the most common fixture (the one with mobilepay), add some yourself and dump an updated fixture with: https://docs.djangoproject.com/en/4.2/ref/django-admin/#dumpdata

_Originally posted by @falkecarlsen in https://github.com/f-klubben/stregsystemet/issues/346#issuecomment-1693325145_


krestenlaust:
I've already pruned some of the irrelevant entries in the fixture. I was considering removing the auth related entries as well, but I'm not certain.

stregsystem/fixtures/testdata-mobilepay.json
```diff
@@ -306,20 +310,910 @@
...
+      "codename": "view_session"
+    }
+  },
   {
     "model": "auth.user",
     "pk": 1,
     "fields": {
-      "password": "pbkdf2_sha256$30000$ivRA8qABWClG$pxi2R4uyotlGpwa66RQuDOXFaO9JElQutMFJqXmnOdI=",
-      "last_login": "2017-03-13T12:41:13.546+00:00",
+      "password": "pbkdf2_sha256$150000$fUAN4hBdKvOp$+1qkLByq1ae/ybzfplIUZnLQvQzHh2NrYLHvzg028Hc=",
```
falkecarlsen, line 1207: Have you changed the test user `tester`'s password? That shouldn't happen.

jonasKjellerup: Without having verified it, the password is likely the same.

The number of rounds of hashing has been increased from 30000 in the original to 1500000 in the new one, resulting in a different hash. This is most likely just an internal change in django since the original fixture was created.

falkecarlsen: Right, now I see `sha256$150000`.

falkecarlsen (approved):
:shipit:

PR #352, opened 2023-09-03
Add Purchase heatmap

### LINKED ISSUE #328: Make github activity heatmap-like statistics for purchases
I like comparing e-penises, therefore we should have a stat on the userinfo page showing a heatmap of purchases distinguished by colour on category perhaps. 

I'm sure a generic lookalike exists.

krestenlaust:
The goal of this PR is to add a heatmap over purchase activity, similar to Githubs activity heatmap, to userpages in Stregsystemet.

Plan:
 - Mix day-cell colors based on what products are purchased on a given day, red=beer, green=energy, blue=soda (example categories).
 - The brightness of a cell is proportional to money spent in relation to the day with most money spent.

Decisions to make:
- Should there be a way to collapse the statistics (and store it on the user)? (to enable privacy for users not using multibuy)

 - ~~Should cell color mixing be based on value sum or product count, in each category?~~
**Have multiple views/modes, one for mixing cells, one for value sum, and one for product count.**

 - ~~Maybe make brightness regard item count as opposed to money, to have more common brightness values as opposed to all kinds of percentages.~~
**Have multiple views/modes.**

 - ~~What page to put the statistics on. This will probably be dependent on how slow the queries are going to be, otherwise I'd prefer it be on a non-obscure page like usermenu (as opposed to statistics-page).~~
**Performance should be okay for it to be included directly in user menu.**
 - ~~Consider what to do with products outside of a category~~
 **Simply disregard them, since it's only relevant to a single of the heatmap modes.**

Resources:
- https://www.alanzucconi.com/2016/01/06/colour-interpolation/

TO-DO:

- [x]  ~~Replace mockup with generated html~~ (Finished by da2d722)
- [x]  ~~Make functions for calculating color based on data~~ (Finished by 8907d1f)
- [x]  ~~Query database for actual data instead of using mockup data~~ (Finished by 485ff25)
- [x]  ~~Finish basic heatmap using database lookup~~ (Finished by 2bb08be)
- [x]  ~~Finish abstraction of heatmap color modes~~ (Finished by 8328017)
- [x]  ~~Improve queries (There's a bunch of similar queries to db)~~ (Finished by 25e3beb. Selects related fields immediately)
- [x]  ~~Need a way to display what products have been bought on a particular day~~ (Finished by d59af25)
- [x]  ~~Add money-sum heatmap color mode (Similar to general, but in relation to money used instead of product count)~~ (Finished by aee0aed)
- [x]  ~~Implements tests (suggestions: empty user, correct date on data)~~ (Finished by accee71)
- [x]  ~~Polish (make more interactive, and possibly interpolate colors instead of just using RGB values)~~ (Finished by 155e227)
- [x]  ~~User evaluation & feedback~~ (Did a few tests with different people, they were able to find the different modes of the heatmap, and were able to successfully navigate the heatmap in terms of dates and weeks. The category view was confusing to the user because the colors are off)
- [x]  ~~Rephrase/remove comments, and improve code~~ (Finished by ad06c7e)
- [x]  ~~Category view doesn't mix colors in the expected way. Awaiting feedback from DVML student~~ (Skipped)
- [x]  ~~(Maybe later?) Add admin settings to control defaults since categories need to be assigned somehow~~ (Not finished :shipit:)

This PR closes #328.

Finished demonstration:
HeatmapDemo


krestenlaust:
The handwritten HTML has been replaced with generated HTML. Mockup data is randomly generated:

Screenshot of current layout

krestenlaust:
For some reason the product instances don't have any categories assigned 🤔 
Debugger view

Even though they clearly are assigned:
Admin view

This is how I retrieve the product instances:
Code


@falkecarlsen have you got any idea why this is happening, or any debugging advice for how to move on ?

falkecarlsen:
You should look into https://docs.djangoproject.com/en/4.2/topics/db/queries/#lookups-that-span-relationships

On mobilepay-fixture I created a category which product pk=1 has, then the following filter works:
`Product.objects.filter(categories__pk=1)`

The Django ORM sometimes seems like it hallucinates completely unrelated predicates into filters, but that's just how it resolves SQL-join operations.

falkecarlsen:
Have you figured out why CI doesn't since [2bb08be](https://github.com/f-klubben/stregsystemet/pull/352/commits/2bb08bef6f1c9f346a926f3219e70cb98a164a75)?

krestenlaust:
I was confused as to it as well. But it looks like it started happening around the same time as this PR was merged https://github.com/f-klubben/stregsystemet/pull/353 🤔 I assumed it was on purpose

krestenlaust:
Updated look:

falkecarlsen:
@krestenlaust be aware that I have dirtied the branch. CI now runs on forks as well, and deps are cached, so turnaround is slightly faster.

krestenlaust:
> @krestenlaust be aware that I have dirtied the branch. CI now runs on forks as well, and deps are cached, so turnaround is slightly faster.

Thanks for heads-up, cool that you got it working!

I plan for this PR to be ready for merging some time this week! Might try to clean-up commit history if I feel fancy

krestenlaust:
> Turns out in Github Action `on: push` declaration only scopes the upstream repo, but `on: pull_request` scopes all PR's across all forks 🤮
> 
> Don't worry about the git history, as we always squash on merging. Try avoiding force pushes also.
> 
> Looking forward to review the feature :fritfit:

Interesting...


Also, I'll keep avoiding force pushes in mind :fritfit:

JakobTopholt:
This is quite a nice feature, but starting the week with a Sunday is indeed haram.

krestenlaust:
@JakobTopholt 
> This is quite a nice feature, but starting the week with a Sunday is indeed haram.

Yes, I'm having doubts about it as well. Another downside of it, is the week numbering doesn't line up properly with Danish week numbers for sunday, so the first displayed cell actually belongs to previous week. This introduces silly bugs like the week number not appearing on the new week on sunday:

(screenshot of heatmap last sunday)


(screenshot of heatmap today)
billede

It's up to @falkecarlsen, I don't have a strong opinion about it. Generally I don't mind the idea of starting the week on a sunday, this also makes wednesday the true halfway point of the week.

falkecarlsen:
I think it's a nice nod to our users that it emulates the Github activity graph, which starts on a Sunday. This also makes the weekday padded by Sat and Sun in top and bottom, and therefore centered.

I see the issue with week-numbers, but then again; do we need or want the week-numbers? Having months for reference seems a good tradeoff:
image

krestenlaust:
Focus points for review:
 - Making use of Q-objects in queries (?)
I don't have any intuition in determining when to use Q-objects, so I've just been using regular queries.
 - Performance
I have been considering performance while developing, but I can't deny that it takes ~300 ms extra to load user-menu page now. I'd really prefer the chart stays on user-menu page, because of my efforts in making it.
 - Category chart doesn't take proportions into consideration and looks a bit crude because of it.
I'll admit. I couldn't figure out a way to make each color component scale according to proportion AND do so inversely, (lower proportions = lighter colors, like the other charts). It feels like a super simple problem, but I've been pulling my hair out for long enough now. *A possible solution to the crude look would be to simply take an average of three (nicer) colors, and not including a color in the average if it isn't represented (e.g. beer and soda day = blue and red average).*

If you're able to boot up a version (If you're TTS isn't too high), it would be great to make sure we're on the same page about everything! I've included a fixture. I've updated it since, so I'll try to reupload it.

> @falkecarlsen I see the issue with week-numbers, but then again; do we need or want the week-numbers? Having months for reference seems a good tradeoff

For now I think I prefer weeks. They're still more precise than months (and they don't require me to make semi-large changes to the code😅)

:fritfit:

jonasKjellerup (commented):
I've added a few minor suggestions. There doesn't seem to be all that much in the way low hanging fruit for performance improvements, but I wouldn't worry all that much about it.

With that said, you could consider trying to reduce the number of round-trips that you are doing to the database. I supsect that for your category lookups, the lookup time probably isn't much longer than the time to establish the connection and build the query. By changing `get_category_objects` to take a list of category names and doing a query akin to this:
```sql
select category_name, product_name
from Categories
inner join Products on Categories.id = Products.category
where category_name in ('category 1', 'category 2')
order by category_name
```
you could do the lookup in one query instead of 6. Not sure what this maps to in Django ORM land, but the general idea is just to reduce the number of lookups you are doing.

Additionally you can add an index on category_name and maybe save a milisecond or two on the lookup.

krestenlaust:
> With that said, you could consider trying to reduce the number of round-trips that you are doing to the database. I supsect that for your category lookups, the lookup time probably isn't much longer than the time to establish the connection and build the query. By changing get_category_objects to take a list of category names and doing a query akin to this:

I've been looking into it, but I haven't been able to make it quite as good as your sql example. Unfortunately, I'm not that versed in Django ORM, and I keep finding the dynamics/behavior of their many-to-many fields (a feature categories are using) ... odd.

I've managed to reduce it from 6 queries to 3 queries, by combining individual category lookups with product.id selections.


I'll continue looking into the other points you mentioned

krestenlaust:
I did some performance improvements since last review, but I haven't been able to optimize it completely (which I'm fine with). I've also adjusted the category chart to a satisfactory state.

stregsystem/templates/stregsystem/purchase_heatmap.html
```diff
@@ -0,0 +1,211 @@
...
+    ]
+
+    window.addEventListener("load", (_) => swapTableType(0));
+
+    function swapTableType(index){
+        const header_title = `Forbrugsoversigt (${heatmap_mode_description[index]})`
+        document.getElementsByClassName("heatmap-header-title")[0].dataset.originaltext = header_title;
+        document.getElementsByClassName("heatmap-header-title")[0].innerText = header_title;
+
+        document.getElementById("heatmap-cell-variable").innerHTML = `.heatmap-cell{--cell-color: var(--cell-color-${index});}`;
```
jonasKjellerup, line 143: Consider using `style.setProperty` instead of updating the contents of a style element. https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty

jonasKjellerup (approved):

stregsystem/templates/stregsystem/purchase_heatmap.html
krestenlaust: I gave a shot implementing it, but I found it challenging, I couldn't find a good way to continue using a class selector when using setProperty. I assume it isn't that big a difference in semantics or performance anyway :)

krestenlaust:
@falkecarlsen can you indicate whether you intend on doing a review

falkecarlsen (approved):
Nice :shipit:

PR #380, opened 2023-10-30
Set cursor to the F-klub logo

Mast3rwaf1z:
This PR makes the cursor in stregsystemet into the logo for the F-klub.

All changes of the cursor also has a fallback to `auto`, meaning if anything fails, the cursor will fall back to the default cursor for where it is.

Changes:
* Added directory `stregsystem/static/stregsystem/cursor`
* Added frame{0-14}.png to `stregsystem/static/stregsystem/cursor`
* Added file `cursor.css` to `stregsystemet/static/stregsystem`, This file has 4 rules:
    * `*` sets the cursor to `stregsystemet/static/stregsystem/cursor/frame14.png` 
    * `input:hover` sets the cursor to a rotating logo when hovering over a button or input field
    * `@-webkit-keyframes cursor` browser compatibility, see below
    * `@keyframes cursor` defines at how far into the animation each frame should be displayed
* added `cursor.css` as file import in `stregsystem/templates/stregsystem/base.html`

*note: i don't care too much whether this gets merged, i'm just familiarizing me with the system*


Mast3rwaf1z:
> Could you post a pic of the cursor on the stregsystem on a 1080p viewport to get a feel for how bad the usability would be?

have a video :)

https://github.com/f-klubben/stregsystemet/assets/32744171/87e3ab1a-c00b-4ddb-ae57-9fa625d5564f

falkecarlsen:
It would be interesting to see how hard the stregmaskine is pushed when having to render animated cursor + 30 bats.

Mast3rwaf1z:
> It would be interesting to see how hard the stregmaskine is pushed when having to render animated cursor + 30 bats.

here, have another video in a slow VM (1024 MB ram, 1 core, no gpu). The server is running on the host system

and its only like 3% more animation anyway :P


https://github.com/f-klubben/stregsystemet/assets/32744171/f12b7722-1290-4d51-ab59-e530bdfdb5f7

falkecarlsen (approved):
:shipit:

PR #382, opened 2023-11-02
Fix preflight header for sale api

LowkeyCoding:
This fixes the issue (https://github.com/f-klubben/fappen/issues/62) in fappen. 

The only real change is allowing for the use of the "content-type" header, which is required for sending JSON to the "sale" API endpoint.


krestenlaust (approved):
LGTM

krestenlaust:
Should I just merge it?

PR #393, opened 2023-11-09
Fix division by zero error in heatmap

krestenlaust:
Forgot to handle the case of a user only buying products which cost 0 F$.


Casper-NS (approved):

jones-g (commented):
I fully endorse this product and/or service

PR #399, opened 2023-12-01
Added white links to the blue background

Murmeldyret:
It is impossible to see the links on the blue background when it is christmas. This change will force all links to be white, when christmas time arrives.


krestenlaust (approved):
LGTM now 👍  thanks for coming up with the css needed!

Mast3rwaf1z (approved):
LGTM, works on my machine

only thing i don't like is how the text in the table under `Brugerinfo`'s lowest row is the same color as the background.
Screenshot_20231204_174925

krestenlaust:
I've opened a new (aptly named) issue for this: #402

PR #401, opened 2023-12-02
Fixed heatmap christmas mode styling

krestenlaust:
Had accidentally specified the color to be black instead of just leaving it as default.


jonasKjellerup (approved):

stregsystem/templates/stregsystem/purchase_heatmap.html
```diff
@@ -67,7 +67,6 @@ <h2 class="heatmap-header-title">Forbrugsoversigt (Antal)</h2>
     .heatmap-label {
         position: relative;
         padding: .125em .5em .125em 0;
-        color: black;
         fill: black;
```
Mast3rwaf1z, line 70: what does fill:black do in this context? changing it in style editor doesn't seem to do anything

Mast3rwaf1z (approved):
looks good, i've attached a screenshot of the style definitely following its parent

Screenshot_20231204_180720

stregsystem/templates/stregsystem/purchase_heatmap.html
krestenlaust: Good catch!

PR #403, opened 2023-12-15
Add none handling in mobilepay-payments

krestenlaust:


stregsystem/management/commands/importmobilepaypayments.py
```diff
@@ -55,8 +55,15 @@ def read_token_storage(self):
         with open(self.tokens_file, 'r') as json_file:
             self.tokens = json.load(json_file)
 
+        if self.tokens is None:
+            self.write_error("read token from storage. 'tokens' is None")
```
falkecarlsen, line 59: Would be great to have the backup token be copied in on errors and either returning or rerunning the import.

If immediately rerunning there **needs** to be a delay sufficient to not trigger API to rate-limit us and preferably also a backoff. If this is not done, we risk burying the server in IO-operations.

falkecarlsen (changes requested):

stregsystem/management/commands/importmobilepaypayments.py
krestenlaust: Are you proposing that we load the `tokens.json.bak` instead when `token.json` can't be loaded?

falkecarlsen: Yes. We never write to tokens.json.bak except the first time that we populate it from the backup in my home-dir.

krestenlaust:
I've implemented a crude solution as I understood it now.

P.s. do you know why black keeps failing, even though I can't find any format suggestions locally @falkecarlsen ? I think it might be whitespace related, I tried fixing it in #408

falkecarlsen (approved):

PR #408, opened 2024-02-22
Attempt fix black formatter

krestenlaust:
At some point, a whitespace difference was introduced somewhere that triggers a black warning on various pull requests now


krestenlaust:
Welp

Mast3rwaf1z (approved):
Running the tests locally still works and it looks like the checks this pr is fixing is completing successfully so it looks good to me

Issue #389, opened 2023-11-04
Update Python version to at least 3.10


krestenlaust:
The current version we're utilizing (3.6) reached [EOL](https://devguide.python.org/versions/) two years ago, having it updated along with the dependencies will make sure the Stregsystem will live long and prosper!

![Picture of python release cycle](https://github.com/f-klubben/stregsystemet/assets/17174333/f697bf46-9238-492b-a8f9-325bbd43392b)

jonasKjellerup:
Since 3.8 will be EOL in < 1 year, would it not make more sense to target a version with a greater remaining support window?

krestenlaust:
The reason I ended up with 3.8 is that the latest version of Pillow requires that version at least, and upgrading in small steps to the common denominator seemed easier. But I agree having a newer version would be better, which version would you suggest?

jonasKjellerup:
Well, considering the fact that Python makes no forward or backwards compatibility guarantees, I don't really see any good rationale to make a choice based on, besides asking ourselves "How soon do we want to do this again?"

We can try to look at what the newest releases of our dependencies should support and use that. However this assumes that we will be updating all of our dependencies. Most of our dependencies are classified as "Python 3.11"  (or newer) on PyPi, but not all. Namely:

- `freezegun` is classified as 3.9. Latest PyPi release is Aug 12, 2022. Latest Github commit Feb 22, 2023.
- `qrcode` is classified as 3.10. Latest PyPi Release is Feb 5, 2023. Latest Github commit Feb 22, 2023.


So we may consider 3.9 as a minimum target, where all our dependencies should "just work" (assuming that they haven't had any breaking changes).

So i see three possible routes:
 - Target the newest release possible and deal with all the issue that come with that.
 - Target the minimum version that should be supported by the newest releases of all our dependencies (3.9).
 - Or lastly, increment vesions until something breaks and stop there. (Although I don't think we are going to make is very far).

With that said, if our goal is to keep Stregsystemet up to date with security support, then the greater issue is that we are using Django 2.2 which went EOL Apr. 1 2022 [(Django support support table)](https://endoflife.date/django). A upgrade to a new major version (3.2 LTS or 4.2 LTS) is more or less guaranteed to break plenty of things.

BoAnd:
@jonasKjellerup what about PR: #409

krestenlaust:
It looks like Freezegun supports 3.12 now

Issue #192, opened 2019-12-02
Upgrade to Django 4.0


VirtualSatai:
Django 3.0 is out and stregsystemet should probably upgrade at some point: https://docs.djangoproject.com/en/3.0/releases/3.0/

Edit: Django 4.0 is out too https://www.djangoproject.com/weblog/2021/dec/07/django-40-released/

PR #409, opened 2024-02-25
Upgrade packages.

BoAnd:
Upgrade packages in requirements:
-  Django==2.2.28 -> Django==5.0.2
- Pillow==8.3.2 -> Pillow==10.2.0
- Coverage==4.4.1 -> Coverage==7.4.3
- pytz==2021.3 -> pytz==2024.1
- freezegun==0.3.15 -> freezegun==1.4.0
- Django-Select2==5.11.1 -> Django-Select2==8.1.2
- django-debug-toolbar==1.11.1 -> django-debug-toolbar==4.3.0
- requests==2.24.0 -> requests==2.31.0
- qrcode==6.1 -> qrcode==7.4.2

The code should run. But I have not test all parts of the stregsystem.


krestenlaust:
( #409 #192 #282 #361 #332  )

Resolves #389

.github/workflows/django.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ["3.6.9", "3.7.17", "3.11.5"] # 3.6.9 is what is running on prod
+        python-version: ["3.10.14", "3.11.9"] # 3.11.9 is what is running on prod
```
krestenlaust, line 17: I'm not sure about what versions we want to have here, so I just picked the one we're updating to and the one before that

treo/settings.py
```diff
@@ -71,7 +72,7 @@
 
 CSRF_COOKIE_SECURE = cfg.getboolean("debug", "CSRF_COOKIE_SECURE")
 CSRF_COOKIE_HTTPONLY = cfg.getboolean("debug", "CSRF_COOKIE_HTTPONLY")
-CSRF_TRUSTED_ORIGINS = ["fappen.fklub.dk"]
+CSRF_TRUSTED_ORIGINS = ["https://fappen.fklub.dk"]
```
krestenlaust, line 75: Just to eliminate any suspicion, this shouldn't break anything with the Fappen, right @LowkeyCoding ?

krestenlaust (approved):
I think it's ready now, the tests run, but I haven't had a thorough run-through of the application yet, just to make sure all untested features still work as well

.github/workflows/django.yml
krestenlaust: Guess it's fine 🤷‍♀️

PR #413, opened 2024-03-01
Fix more errors

krestenlaust:
Errors that were introduced when changing the `Member`-model.


krestenlaust:
Mistankenly made PR to parent repo instead of fork

PR #416, opened 2024-03-12
Implement new Mobilepay reporting API

### LINKED ISSUE #407: MobilePay Reporting API going out of support
MobilePay is in the midst of doing a transition to their new Vipps MobilePay platform. This means that the current implementation for fetching transactions will stop working come ~mid-2024.

The migration should in theory be fairly straightforward, but since we are still on the already deprecated V1 of the original API it may be more substantial, as the migration guides assume that we are on V3.

https://developer.mobilepay.dk/docs/reporting/transition-to-one-platform

Reporttimeline-055fcc36eacd2e88ad6d036b34fc7d47

Migration information: https://developer.vippsmobilepay.com/docs/mp-migration-guide/reporting/
Documentation for the replacement API: https://developer.vippsmobilepay.com/docs/APIs/report-api/

krestenlaust:
Resolves #407 

New API endpoint: https://api.vipps.no/

Resources:
 - Migration - https://developer.vippsmobilepay.com/docs/mp-migration-guide/reporting/
 - API key details - https://developer.vippsmobilepay.com/docs/partner/partner-keys/
 - Report API - https://developer.vippsmobilepay.com/api/report
 - Access Token API - https://developer.vippsmobilepay.com/docs/APIs/access-token-api/

### DIFF
--- a/stregsystem/management/commands/importmobilepaypayments.py
+++ b/stregsystem/management/commands/importmobilepaypayments.py
@@ -23,6 +23,7 @@ class Command(BaseCommand):
     tokens_file = (Path(__file__).parent / 'tokens.json').as_posix()
     tokens_file_backup = (Path(__file__).parent / 'tokens.json.bak').as_posix()
     tokens = None
+    # Cutoff for when this iteration of the Mobilepay-API (Vipps) is deployed
     manual_cutoff_date = date(2024, 4, 9)
     myshop_number = 90601
 


krestenlaust:
Stale until we get a response as to whether or not API-keys will be available to us

krestenlaust:
Notable difference:
VippsMobilepay doc

There's no transaction ID in the new API, so I'll be using PSP-reference as that seems like it's closest to it. It's only used for recognizing duplicates anyway

stregsystem/management/commands/importmobilepaypayments.py
```diff
@@ -16,12 +17,14 @@
...
     help = 'Imports the latest payments from MobilePay'
 
-    api_endpoint = 'https://api.mobilepay.dk'
+    api_endpoint = 'https://api.vipps.no'
     # Saves secret tokens to the file "tokens.json" right next to this file.
     # Important to use a separate file since the tokens can change and is thus not suitable for django settings.
     tokens_file = (Path(__file__).parent / 'tokens.json').as_posix()
     tokens_file_backup = (Path(__file__).parent / 'tokens.json.bak').as_posix()
     tokens = None
+    manual_cutoff_date = date(2024, 4, 9)
```
krestenlaust, line 26: Is this an appropriate way to prevent registering payments twice?

stregsystem/management/commands/importmobilepaypayments.py
```diff
@@ -16,12 +17,14 @@
...
+    myshop_number = 90601
```
krestenlaust, line 27: Hardcoded myshop number 🤔  Should I move it into the tokens-file?

krestenlaust (commented):
I'm not sure how to test it, I've only been testing the code besides in the interpreter

krestenlaust:
I has been tested in a local setup now, and it imported items successfully

krestenlaust:
The prod tokens are located in my home folder on the server

stregsystem/management/commands/importmobilepaypayments.py
BoAnd: I think it is fine. 
But maybe someone should check there is not missing transactions between stregsystem adn mobilepay.

stregsystem/management/commands/importmobilepaypayments.py
BoAnd: I think that is fine.

stregsystem/management/commands/importmobilepaypayments.py
```diff
@@ -73,68 +76,85 @@ def update_token_storage(self):
...
         }
-        response = requests.post(url, data=payload)
+
+        authorization_string = '{}:{}'.format(self.tokens['client_id'], self.tokens['client_secret'])
+        authorization_base64 = b64encode(authorization_string.encode("ascii")).decode("ascii")
+        headers = {
+            'authorization': 'Basic {}'.format(authorization_base64),
+        }
+
+        response = requests.post(url, data=payload, headers=headers)
```
BoAnd, line 91: Maybe I would have used HTTPBasicAuth instead here. Would look like something like this. It is a small thing. 
```
requests.post(url, data=payload, auth=HTTPBasicAuth(self.tokens['client_id'], self.tokens['client_secret']))
```

BoAnd (approved):
I think it look fine.

stregsystem/management/commands/importmobilepaypayments.py
krestenlaust: Much better, thanks!

stregsystem/management/commands/importmobilepaypayments.py
krestenlaust: Yes, I'm more afraid of there being double transactions (people having more money than they should), I planned the following issue #417. But I'll probably end up simply pulling all manual transactions from the database and comparing them to a CSV from MobilePay (simply summing each user's transactions from each and account for the discrepancy)

stregsystem/management/commands/importmobilepaypayments.py
krestenlaust: To have it moved into the tokens-file or have it stay the way it is?

BoAnd: Sorry. I think it is fine to have it here. 
We are not changing it between builds and it is not a secret.

krestenlaust: Thanks for the clarification! And thanks for reviewing!

stregsystem/management/commands/importmobilepaypayments.py
falkecarlsen, line 26:
```suggestion
    # Cutoff for when this iteration of the Mobilepay-API (Vipps) is deployed
    manual_cutoff_date = date(2024, 4, 9)
```

For future reference

falkecarlsen: Should be fine if API is verified to work initially when deployed and no duplicate payments are registered during the transition period.

stregsystem/management/commands/importmobilepaypayments.py
```diff
@@ -73,68 +76,81 @@ def update_token_storage(self):
...
-        current_time = datetime.now(timezone.utc)
+    def get_transactions(self, transaction_date: date):
+        ledger_date = transaction_date.strftime('%Y-%m-%d')
+
+        url = f"{self.api_endpoint}/report/v2/ledgers/{self.tokens['ledger_id']}/funds/dates/{ledger_date}"
+
         params = {
-            'from': self.format_datetime(current_time - timedelta(days=self.days_back)),
-            'to': self.format_datetime(current_time),
+            'includeGDPRSensitiveData': "true",
```
falkecarlsen, line 105: Are we getting names and phone-numbers now? That'll help matching as I wrote about in #204 somewhere.

falkecarlsen (approved):
LGTM with @BoAnd suggestions.

As you are aware, be careful with duplicate payments during deployment. You could deploy outside business-hours.

stregsystem/management/commands/importmobilepaypayments.py
krestenlaust: We get full names and partial phone-numbers, assuming the name isn't easy to change in the app (somehow tied to MitID), so I guess it could be used to match members with unique name and last 4 digits. A shame we don't get the whole phone-number 🤔 
```
{
    "pspReference": "32212390715",
    "time": "2024-04-05T07:19:26.528092Z",
    "ledgerDate": "2024-04-05",
    "entryType": "capture",
    "reference": "10113143347",
    "currency": "DKK",
    "amount": 20000,
    "recipientHandle": "DK:90601",
    "balanceAfter": 110000,
    "balanceBefore": 90000,
    "name": "Jakob Topper",
    "maskedPhoneNo": "xxxx 1234",
    "message": "Topper"
}
```

falkecarlsen: Name + 4 digits of phone number is pretty much uniquely identifying but requires us to record phone number on our fembers as well, which is a hassle to enroll existing members in.

The name can be a factor in matching, but it shouldn't be mandatory as a fember might for some reason or other want to top off another fembers account.

falkecarlsen: And yes, the name *is* the legal name of the payee: https://mobilepay.dk/hjaelp/mobilepay-til-private/brug/basisfunktioner/oversigt/hvorfor-bliver-mit-fulde-navn-vist

krestenlaust: Another thing to note regarding name, is that the current configuration only allows, respectively, 20 and 30 characters for firstname and lastname, which isn't enough to match a full legal name for many, and you'd also have to enroll people into providing their legal names as well.

The closest to a use-case I see is suggesting members, on unset payments, based on their names if it matches a full name. If we had access to full phone-number, I'd be prepared to enroll existing users if it meant their payments could be registered without a comment.

PR #420, opened 2024-04-07
Document API with OpenAPI spec (API 1.1)

krestenlaust:
In order to make the use of the API more welcoming, the API should be formally described as an OpenAPI specification.

Endpoints:
- [x] api/member
- [x] api/member/balance
- [x] api/member/get_id
- [x] api/member/active
- [x] api/member/sales
- [x] api/member/payment/qr
- [x] api/products/named_products
- [x] api/products/active_products
- [x] api/products/category_mappings
- [x] api/sale


openapi/openapi.yaml
```diff
@@ -0,0 +1,566 @@
...
+        '200':
+          $ref: '#/components/responses/MemberFound_Active'
+        '400':
+          $ref: '#/components/responses/MemberIdParameter_BadResponse'
+  /api/member/sales:
+    get:
+      tags:
+        - Member
+      summary: Get member sales
+      description: Gets a list of a member's sales.
```
AshesOfEther, line 71: Purchases might be a better term? I know this is not the term used by the API, but it is clearer. Members don't have sales; they don't sell stuff.
```suggestion
      description: Gets a list of a member's purchases.
```

openapi/openapi.yaml
```diff
@@ -0,0 +1,566 @@
...
+          $ref: '#/components/responses/MemberUsernameParameter_BadResponse'
+  /api/member/payment/qr:
+    get:
+      tags:
+        - Member
+      summary: Get payment QR code
+      description: Returns a QR code for payment for a member.
+      operationId: get_payment_qr
+      parameters:
+        - $ref: '#/components/parameters/username'
```
AshesOfEther, line 102: This parameter is named `member` in the API, not `username`.

openapi/openapi.yaml
```diff
@@ -0,0 +1,566 @@
...
+      tags:
+        - Sale
+      summary: Perform a sale using a buy-string
+      description: something.
+      operationId: sale
+      responses:
+        '200':
+          $ref: '#/components/responses/SaleSuccess'
+        '400':
+          $ref: '#/components/responses/Member_RoomIdParameter_BadResponse'
```
AshesOfEther, line 154: This is missing parameters and description.

AshesOfEther (commented):
Looks mostly good. I suggest using the sample instance of [Swagger UI](https://petstore.swagger.io/) to check that it's right. Perhaps we could publish that on a docs site of our own? That's out of scope for this PR though.

openapi/openapi.yaml
krestenlaust: Agree, I've "set mig blind" on it, makes good sense!

krestenlaust: I think the original reasoning for the naming of the API was that it was oriented for use by FIT personnel. But the new API is targeting fembers, so it makes sense to change it

openapi/openapi.yaml
krestenlaust: Resolved in [21aaeb1](https://github.com/f-klubben/stregsystemet/pull/421/commits/21aaeb10954bf462cf88df8bc02a61190703735e)

krestenlaust: Good catch!

openapi/openapi.yaml
krestenlaust: Done!

AshesOfEther (approved):

krestenlaust:
> Looks mostly good. I suggest using the sample instance of [Swagger UI](https://petstore.swagger.io/) to check that it's right. Perhaps we could publish that on a docs site of our own? That's out of scope for this PR though.

It would be very nice to have the docs site, it could probably be done through github if not on our own server

PR #421, opened 2024-04-07
Improve API (API 1.1)

krestenlaust:
#420 aims to document the API. This PR addresses inconsistencies found during that documentation effort. Both in implementation and in the actual API.

It increments the API versioning to 1.1


krestenlaust:
QR-endpoint apparently uses Stregdollars as opposed to Stregøre. I feel like streamlining this is too big a change.
The QR-endpoint also takes a `member` parameter, which will probably be renamed to `username` to streamline with other Member API-methods. (`member` will probably still be supported)

krestenlaust:
Changed response text on error to this format

krestenlaust:
Changed format for Category_mapping endpoint output:
`{"1": [], "2": [], "3": [], "13": [], "32": [[6, "kaffe"]], "33": [], "34": []}` -> `{"1": [], "2": [], "3": [], "13": [], "32": [{"category_id": 6, "category_name": "kaffe"}], "33": [], "34": []}`

Much more verbose, but it's more like the other endpoints. Considering changing it to not return products which don't have any mapped categories

krestenlaust:
I've decided to leave the sales API as is

stregsystem/urls.py
```diff
@@ -33,14 +33,14 @@
...
-    re_path(r'^api/member$', views.get_user_info, name="get_user_transactions"),
-    re_path(r'^api/products/named_products$', views.dump_named_items, name="named_products"),
-    re_path(r'^api/products/active_products$', views.dump_active_items, name="active_products"),
+    re_path(r'^api/member/payment/qr$', views.get_payment_qr, name="get_payment_qr"),
+    re_path(r'^api/member/active$', views.get_member_active, name="get_member_active"),
+    re_path(r'^api/member/sales$', views.get_member_sales, name="get_member_sales"),
+    re_path(r'^api/member/get_id$', views.get_member_id, name="get_member_id"),
+    re_path(r'^api/member/balance$', views.get_member_balance, name="get_member_balance"),
+    re_path(r'^api/member$', views.get_member_info, name="get_member_info"),
+    re_path(r'^api/products/named_products$', views.dump_named_products, name="dump_named_products"),
```
Mast3rwaf1z, line 42: why not `get_named_products` and `get_active_products`? this change seems related to making better function names, so why not follow what is above?

stregsystem/views.py
```diff
@@ -563,18 +565,18 @@ def signup_tool(request):
...
         return HttpResponseBadRequest("Invalid input for MobilePay QR code generation")
 
-    query = {'phone': '90601', 'comment': form.cleaned_data.get('member')}
-
-    if form.cleaned_data.get("amount") is not None:
-        query['amount'] = form.cleaned_data.get("amount")
+    username = form.cleaned_data.get('username')
+    amount = form.cleaned_data.get('amount')
 
-    data = 'mobilepay://send?{}'.format(urllib.parse.urlencode(query))
```
Mast3rwaf1z, line 576: remove urllib.parse import

stregsystem/views.py
```diff
@@ -620,78 +622,85 @@ def signup_status(request, signup_id):
...
     username = request.GET.get('username') or None
     if username is None:
-        return HttpResponseBadRequest("Missing username")
+        return HttpResponseBadRequest("Parameter missing: username")
+
     try:
         member = Member.objects.get(username=username)
     except Member.DoesNotExist:
-        return HttpResponseBadRequest("Invalid username")
+        return HttpResponseBadRequest("Member not found")
```
Mast3rwaf1z, line 658: The rest of these error messages are prefixed with something like `Parameter missing:` but not this one

Edit: i see more that aren't prefixed as well, i think it would be nice to have them all prefixed like that

stregsystem/forms.py
```diff
@@ -72,7 +72,7 @@ def __init__(self, *args, **kwargs):
 
 class QRPaymentForm(forms.Form):
     member = forms.CharField(max_length=16)
-    amount = forms.DecimalField(min_value=50, decimal_places=2, required=False)
+    amount = forms.DecimalField(min_value=0.01, decimal_places=2, required=False)
```
Mast3rwaf1z, line 75: I don't see the point in this change? don't we disallow payments below 50?

Mast3rwaf1z (changes requested):
a few small changes, but looks nice :+1:

stregsystem/forms.py
Mast3rwaf1z: also when i send a request to `http://localhost:8000/api/member/payment/qr?member=tester&amount=0.01` it fails, while `amount=0.02` doesn't

stregsystem/views.py
krestenlaust: This should be fixed by 40e878f

stregsystem/forms.py
krestenlaust: Changed in cdaff44

stregsystem/views.py
krestenlaust: Fixed by dba5abd

stregsystem/urls.py
krestenlaust: I've updated the methods by 3c9f739
I've tried changing the naming convention further by f0f5b50

PR #423, opened 2024-04-10
Implement new MobilePay Feed API

krestenlaust:
Excerpt from Mobilepay documentation:
billede
As it turns out, it isn't possible to get payments the same day using the API implemented in #416, so to reach the same service-level as last before the Vipps-Mobilepay fusion, we'll have to implement yet another API.

https://developer.vippsmobilepay.com/api/report/#tag/reportv2ledgers/paths/~1report~1v2~1ledgers~1%7BledgerId%7D~1%7Btopic%7D~1dates~1%7BledgerDate%7D/get

### DIFF
--- a/stregsystem/management/commands/importmobilepaypayments.py
+++ b/stregsystem/management/commands/importmobilepaypayments.py
@@ -197,6 +197,12 @@ class Command(BaseCommand):
         if transaction['entryType'] != 'capture':
             return
 
+        payment_datetime = parse_datetime(transaction['time'])
+
+        if payment_datetime.date() < self.manual_cutoff_date:
+            self.write_debug(f'Skipping transaction because it is before payment cutoff date {payment_datetime}')
+            return
+
         trans_id = transaction['pspReference']
 
         if MobilePayment.objects.filter(transaction_id=trans_id).exists():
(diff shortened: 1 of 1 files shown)


jones-g (approved):
My simple mind has looked over the changes and concluded it looks good.

krestenlaust:
@jones-g I must say that I _love_ the enthusiasm, but I haven't really gotten around to implementing it just yet 🤔 

billede

stregsystem/management/commands/importmobilepaypayments.py
```diff
@@ -112,6 +118,52 @@ def get_transactions(self, transaction_date: date):
...
+            transactions.extend(res['items'])
+
+            try_later = res['tryLater'] == "true"
+
+            if try_later:
+                break
+
+            cursor = res['cursor']
+
+            # Note: Since MobilePay API doesn't return 'hasMore' like the docs says it does.
```
BoAnd, line 142: It says in the response schema that it is only present on the ```.../dates/..``` endpoint. I think they are trolling by showing it in the response sample for fetch report by feed :trollface:

BoAnd:
From the awesome vipps api documentation :vomiting_face:. I think it should work

BoAnd (approved):
From the awesome vipps api documentation :vomiting_face: . I think it should work

stregsystem/management/commands/importmobilepaypayments.py
krestenlaust: oh god 😵‍💫 , I had a hard time wrapping head around why they did that

PR #427, opened 2024-04-22
Fix Buy button greyed out when navigating back from userrmenu

### LINKED ISSUE #356: Buy butten greyed out when navigating back from usermenu
When using browser (Firefox 117) back-button, the purchase button is greyed out. It's most likely something to do with caching and forms-submitting not cooperating properly.

Greyed out button:
Screenshot of purchase button being greyed out

krestenlaust:
Fixes #356

This adds an event listener to the index-page which reenables the buy button on 'pageshow'. Credit to @BoAnd for the fix in #356

### DIFF
--- a/stregsystem/templates/stregsystem/index.html
+++ b/stregsystem/templates/stregsystem/index.html
@@ -98,6 +98,11 @@
 
 <div style="clear: both;">&nbsp;</div>
 
+<script>
+  window.addEventListener("pageshow", (event) => {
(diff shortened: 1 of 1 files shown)


BoAnd (approved):
I think this should work fine. 
What I can see there is full support in all common browser https://developer.mozilla.org/en-US/docs/Web/API/Window/pageshow_event#browser_compatibility

PR #428, opened 2024-04-22
Add tokens sample file

krestenlaust:
With anonymized/randomized values, it shows how the file should be populated.


krestenlaust:
Wrong checkout branch

PR #429, opened 2024-04-22
Vipps API - Sample tokens file

krestenlaust:
Adds a file with sample key values for the Vipps Accounting API. This will help show where the tokens file should be and what it should look like


stregsystem/management/commands/tokens.example.json
```diff
@@ -0,0 +1,4 @@
+{
+  "client_id": "8f51e573-afcc-429d-7b6e-09aacc8f0e86",
+  "client_secret": "YpL8msPGbCKqdJwRnKxUtVb"
```
TobiasSN, line 3: This looks like a real ID and secret, which does make it a bit confusing, because someone might think it's real and report it or try to use it. Maybe something more obviously not real?
```suggestion
  "client_id": "<client id here>",
  "client_secret": "<client secret here>"
```

TobiasSN (commented):
Should hold on #430, and move to the new location if it's merged

stregsystem/management/commands/tokens.example.json
krestenlaust: What do you think of this compromise? I agree that it might be confusing, but I also appreciate having an idea of what kind of API key I'm looking for
```suggestion
  "client_id": "<client id here>", // e.g. 8f51e573-afcc-429d-7b6e-09aacc8f0e86
  "client_secret": "<client secret here>" // e.g. YpL8msPGbCKqdJwRnKxUtVb
```

P.s. I get that "strictly-speaking" JSON doesn't support comments

TobiasSN: Still implicit that they're not real. How about this? To be honest this is not the biggest deal, I won't block the PR on this :sweat_smile: 
```suggestion
  // These are example values. Replace them with real ones and remove this comment
  "client_id": "8f51e573-afcc-429d-7b6e-09aacc8f0e86",
  "client_secret": "YpL8msPGbCKqdJwRnKxUtVb"
```

krestenlaust: You've convinced me to add a disclaimer

PR #430, opened 2024-04-22
Move Vipps API into separate file

krestenlaust:
This is done to make it more intuitive where to find the API implementation.

Major changes:
- `stregsystem/management/commands/tokens.json` -> `stregsystem/vipps-tokens.json`
- Moved API implementation into separate file to give implementation meaningful name (`AccountingAPI`)

Other:
- No significant increase in size of codebase (helped by removing shorthand methods for logging)

I've tested it locally, and it still imports payments without any issues

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -8,4 +8,5 @@ omit =
         */migrations/*
         treo/wsgi.py
         */apps.py
+        stregsystem/vipps_api.py
 source = .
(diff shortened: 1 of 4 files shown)


krestenlaust:
#429 should be changed if this is merged first

BoAnd (approved):
LGTM. 
You could maybe add vipps_api.py to the codecov ignore list (we are talking with a third-party api in the file)

krestenlaust:
> LGTM. You could maybe add vipps_api.py to the codecov ignore list (we are talking with a third-party api in the file)

Good point!

PR #432, opened 2024-04-23
Fix multibuy problem.

BoAnd:
This should fix the problem when buying more then one product. But only one of each. 
```
{username} 14:1 47:1
```


krestenlaust (approved):
Looks reasonable

PR #433, opened 2024-04-23
Better purchase format.

BoAnd:
@krestenlaust was it something like this?
image


BoAnd:
Yes, we can remove the regex (stupid me). 
I need to test it some more.

krestenlaust (approved):
LGTM. Tests run, and it works on my machine. I just removed the unused import.

PR #434, opened 2024-04-25
Feature: Mail/download userdata

Mast3rwaf1z:
The idea is to allow users to download all their user data, either locally or over email

To avoid people downloading to strandvejen i should probably also add some way to hide the download button there, or omit it all together...


Mast3rwaf1z:
@krestenlaust i think this is a working version, but i have no way of testing whether the mails are being sent properly, do you have an idea?

krestenlaust:
I think I'll try to do a review just now, and then we can find a way to test it afterwards. But haven't you been able to set up a local mailserver before?

Mast3rwaf1z:
i'll see if i can find something to test with, but the main problem is whether i'd be spamming anything, so i figured i'd ask if you used something to test mails from stregsystemet

stregsystem/templates/stregsystem/sent_csv_mail.html
```diff
@@ -0,0 +1,19 @@
...
+    <b>payments.csv</b> indeholder alle indbetalinger du har gjort gennem Mobilepay.<br>
+    <b>userdata.csv</b> indeholder alt data i den nederste tabel på forrige side.<br><br>
+
+    Blev der sendt en mail? <b>{{mail_sent_text}}</b>
+    
+    <h3><a href="/{{room.id}}/user/{{member.id}}">Tilbage til forrige side</a></h3>
+
+</center>
+
+{% endblock %}
```
krestenlaust, line 19: I'd imagined that the functionality could be incoperated into the existing "User-info" page, maybe in the bottom of user-info, (Instead of adding a whole other page)

Mast3rwaf1z:
https://github.com/mailhog/MailHog seems good enough, i'll test it some more before putting it ready for review

stregsystem/urls.py
```diff
@@ -32,6 +32,7 @@
     re_path(r'^api/member/payment/qr$', views.qr_payment, name="payment_qr"),
     re_path(r'^api/member/active$', views.check_user_active, name="active_member"),
     re_path(r'^api/member/sales$', views.get_user_sales, name="get_user_sales"),
+    re_path(r'^(?P<room_id>\d+)/send_csv_mail/(?P<member_id>\d+)/$', views.send_csv_mail, name="send_csv_mail"),
```
krestenlaust, line 35: If this endpoint is still needed, I think it should be moved above the API call section of urls

stregsystem/templates/stregsystem/sent_csv_mail.html
```diff
@@ -0,0 +1,19 @@
+{% extends "stregsystem/base.html" %}
+
+{% block content %}
+<center>
+    <h2>Ærede fember ({{member.username}}), hvis du ikke har efterspurgt data de sidste 10 dage, skulle du gerne have modtaget en email.</h2>
```
krestenlaust, line 5: I was thinking more in the lines of 10 minutes rather than 10 days

krestenlaust (changes requested):
Nice work! I've added some comments, if there's anything you'd like expanded or so, we can just find a day to look at it at the same time :D

krestenlaust:
> https://github.com/mailhog/MailHog seems good enough, i'll test it some more before putting it ready for review

No worries, I'll make sure to review again when you're ready 👌

stregsystem/mail.py
```diff
@@ -46,3 +48,48 @@ def send_template_mail(member, target_template: str, context: dict, subject: str
...
+        attachment = MIMEApplication(csv, Name=f"{name}.csv")
+        attachment['Content-Disposition'] = f'attachment; filename="{name}.csv"'
+        msg.attach(attachment)
+
+    try:
+        smtpObj = smtplib.SMTP('localhost', 25)
+        smtpObj.sendmail('treo@fklub.dk', member.email, msg.as_string())
+    except Exception as e:
+        logger.error(str(e))
+    return True
```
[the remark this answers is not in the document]
Mast3rwaf1z: i was thinking the same, but it is a draft pr after all :)

stregsystem/templates/stregsystem/sent_csv_mail.html
Mast3rwaf1z: the problem was that i didn't want to load the entire tables every time a user would access the page

Mast3rwaf1z:
it works using `mailhog -smtp-bind-addr 0.0.0.0:25`, but i'd like to have a new column in the member table called like `last_data_request` but adding it to the member object in models.py doesn't work, how do i go about adding a column?

stregsystem/mail.py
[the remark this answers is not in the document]
krestenlaust: No worries!

stregsystem/templates/stregsystem/sent_csv_mail.html
krestenlaust: Hmm I get that argument, it would probably be easiest to simply have it that way. Otherwise you should do like with purchases (the same page is returned with purchase info) and return to the same page with arguments

stregsystem/mail.py
```diff
@@ -30,7 +32,42 @@ def send_payment_mail(member, amount, mobilepay_comment):
...
+        "send_csv.html",
+        {**vars(member), "fember": member.username},
+        f'{member.username} has requested their user data!',
+        {"sales.csv": sales_csv, "payments.csv": payments_csv, "userdata.csv": userdata_csv},
+    )
+    member.save()
+    return True
+
+
+def send_template_mail(member, target_template: str, context: dict, subject: str, attachments: dict = {}):
```
krestenlaust, line 70: Nice work updating the send template mail 👌

stregsystem/models.py
```diff
@@ -172,6 +172,7 @@ class Member(models.Model):  # id automatisk...
     balance = models.IntegerField(default=0)  # hvor mange oerer vedkommende har til gode
     undo_count = models.IntegerField(default=0)  # for 'undos' i alt
     notes = models.TextField(blank=True)
+    requested_data_time = models.DateTimeField(blank=True, null=True, default=timezone.datetime.fromtimestamp(0))
```
krestenlaust, line 175: I don't think this change is worth updating the member model, though nice work anyhow with your implementation draft. I think it's fine to just have it in an transient dictionary like you did previously. Or something like that. I'm not entirely sure, maybe we should just discuss this face-to-face some time

stregsystem/mail.py
```diff
@@ -30,7 +32,42 @@ def send_payment_mail(member, amount, mobilepay_comment):
...
+    # haha linting as i have no idea how django works otherwise
+    from .models import Payment, Sale
+
+    sales: list[Sale] = member.sale_set.order_by("timestamp")
+    payments: list[Payment] = member.payment_set.order_by("timestamp")
+
+    sales_csv = "Name, Price, Timestamp\n"
+    sales_csv += "\n".join([f'{sale.product.name},{sale.price},{sale.timestamp}' for sale in sales])
+    payments_csv = "Timestamp, Amount\n"
+    payments_csv += "\n".join([f'{payment.timestamp},{payment.amount}' for payment in payments])
```
krestenlaust, line 55: It could be cool to also tell which payments are MobilePay automated, even though it's an implementation detail on our part.

There's a model called MobilePayment you can look into for this, I'm not entirely sure how it works

stregsystem/templates/mail/send_csv.html
```diff
@@ -0,0 +1,7 @@
+<html>
+    <body>
+        Ærede fember - {{fember}}
+
+        Hermed har du din bruger data i stregsystemet i form af csv filer.
```
krestenlaust, line 5:
```suggestion
        Ærede fember - {{fember}}

        Du har anmodet om at få tilsendt al den brugerdata vi har opbevaret af din bruger. Din data er vedhæftet som CSV-filer.
        
        *Indsæt hilsen fra de andre mails*
        ---
        *Write in english*
```

stregsystem/views.py
```diff
@@ -242,6 +242,17 @@ def menu_userinfo(request, room_id, member_id):
     return render(request, 'stregsystem/menu_userinfo.html', locals())
 
 
+def send_csv_mail(request, room_id, member_id):
+    from .mail import send_csv_mail as _send_csv_mail
+
+    room = Room.objects.get(pk=room_id)
+    member = Member.objects.get(pk=member_id, active=True)
+
+    mail_sent_text = "Mail sendt!" if _send_csv_mail(member) else "Ingen mail sendt."
```
krestenlaust, line 251: Maybe pass a boolean instead, and have this condition in the template instead.

krestenlaust (changes requested):
Overall great improvement and awesome work

Is it possible for you to rename references from CSV to something like userdata. "send_csv.html" to "send_userdata.html", "sent_userdata_mail.html" etc. As CSV is an implementation detail.

If all these requests are too much work, then please let me know, and I'll give a hand, as this is an awesome idea.

stregsystem/mail.py
```diff
@@ -30,7 +32,42 @@ def send_payment_mail(member, amount, mobilepay_comment):
...
+    userdata_csv = "Id, Name, First name, Last name, Email, Registration year\n"
+    userdata_csv += f"{member.id},{member.username},{member.firstname},{member.lastname},{member.email},{member.year}"
```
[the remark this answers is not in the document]
Mast3rwaf1z: I disagree, why should we generate additional IO by using this library to write a file? This library looks to be for creating files, atm i'm just making a bytestream from memory, while if i used this library i would have to make a csv writer object, write the database into a file with this object, read the file, convert it to a csv reader, convert that to a string and then i can attach it

seems like a ton of unnecessary IO, but i might be wrong, its just a quick google search

as for spaces, yes that was a mistake :P

stregsystem/models.py
Mast3rwaf1z: what was there before with the dict didn't work, the state was not saved between loads of the page, hence i wanted to store it in the database

stregsystem/mail.py
Mast3rwaf1z: it should work now, although i don't know how to test it, as it seems the db has no payment objects in any of the testdata?

stregsystem/templates/stregsystem/sent_csv_mail.html
Mast3rwaf1z: i'll close this one as it sounds like its good

stregsystem/mail.py
```diff
@@ -30,7 +32,42 @@ def send_payment_mail(member, amount, mobilepay_comment):
     )
 
 
-def send_template_mail(member, target_template: str, context: dict, subject: str):
+def send_csv_mail(member):
+    now = timezone.now()
+    if member.requested_data_time is not None:
+        ten_minutes_ago = now - timezone.timedelta(minutes=10)
+        if member.requested_data_time > ten_minutes_ago:
```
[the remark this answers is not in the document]
Mast3rwaf1z: so you're thinking adding it to a config file? makes sense

stregsystem/templates/stregsystem/sent_userdata.html
```diff
@@ -0,0 +1,31 @@
+{% extends "stregsystem/base.html" %}
+
+{% block content %}
+<center>
+    <h2>Ærede fember ({{member.username}}), hvis du ikke har efterspurgt data de sidste 5 minutter, skulle du gerne have modtaget en email.</h2>
```
krestenlaust, line 5: You already know whether a mail was sent (`mail_sent`), so I think this text should just adapt to that. I get the joking tone, but I think the 'honored fember' part is a bit too modest. Maybe just start the text with "Du skulle gerne have modtaget en mail..." somthing along those lines depending on whether the mail was sent or not

stregsystem/templates/stregsystem/sent_userdata.html
```diff
@@ -0,0 +1,31 @@
...
+            <th><b>payments.csv</b></th> 
+            <th>indeholder alle indbetalinger du har gjort gennem Mobilepay.</th>
+        </tr>
+        <tr>
+            <th><b>userdata.csv</b></th> 
+            <th>indeholder alt data i den nederste tabel på forrige side.</th>
+        </tr>
+    </table><br>
+
+    Sidst du fik sendt mail: <b>{{sent_time}}</b><br>
```
krestenlaust, line 23: Since this isn't stored in the model anymore, we can't really expect it to know the time. So we should probably only serve this time, if the mail isn't sent, "Du fik tilsendt en mail 2 minutter siden, vent til 5 minutter" or something along those lines.

stregsystem/urls.py
```diff
@@ -32,6 +32,7 @@
     re_path(r'^api/member/payment/qr$', views.qr_payment, name="payment_qr"),
     re_path(r'^api/member/active$', views.check_user_active, name="active_member"),
     re_path(r'^api/member/sales$', views.get_user_sales, name="get_user_sales"),
+    re_path(r'^(?P<room_id>\d+)/send_csv_mail/(?P<member_id>\d+)/$', views.send_userdata, name="send_csv_mail"),
```
krestenlaust, line 35: This still needs to be moved

krestenlaust (changes requested):
These are the final changes I can think of. Should be ready to go after this!

Mast3rwaf1z:
just quickly went over your review in my lunch break :)

could you please take a look at the two comments i haven't closed further up? @krestenlaust

krestenlaust (approved):
Nice work! LGTM

@falkecarlsen this issue is related to GDPR and serving userdata. Have you got an opinion regarding the implementation or otherwise ?

BoAnd:
LGTM

BoAnd (approved):

krestenlaust:
@Mast3rwaf1z you may merge whenever :D If you at some point in the future feel like doing more on this feature, then adding more tests would be cool

PR #444, opened 2024-05-02
Readd refund action

### LINKED ISSUE #418: Refund sale
Option to refund a sale (delete it and add the amount back into the account) and return a payment (delete it and subtract the amount from the account)

krestenlaust:
I deindented this code and inadvertendly disabled the feature

Closes #418


JakobTopholt (approved):
I can't really see why this wouldn't work, so Imma approve it without testing it.

In short, LGTM.

PR #451, opened 2024-05-14
Upgrade Python to 3.11.5 and upgrade dependencies

### LINKED ISSUE #389: Update Python version to at least 3.10

krestenlaust:
Only upgraded Django to major 3, since 4 and 5 had issues with tests.

The one upgrade PR to rule them all ( #409 #192 #282 #361 #332  )

Resolves #389


krestenlaust:
Closed in favor of existing update PR

PR #453, opened 2024-05-15
[Bug fix] Sanitize csv mails

Mast3rwaf1z:
in progress fix to unsanitized values in csv files, there are two issues atm

* utf-8 values like æøå
* commas


krestenlaust (commented):
What about products with " in the name

Mast3rwaf1z:
> What about products with " in the name

Hmm good point, i'll just have a google to see how others solve that as well, no way we have to consider everything :P 

I have tested it using mailhog like in #434, `mailhog --smtp-bind-addr 0.0.0.0:25` and this python script:
```py
import csv
with open("Downloads/sales.csv") as file:
    reader = csv.reader(file)
    for row in reader:
        print(row)
```
the two issues i've mentioned is gone now at least

Mast3rwaf1z:
> What about products with " in the name

this should be covered now

Mast3rwaf1z:
I have an alternative idea that i just tested, instead of having the rows_to_csv function we have now, we can go with your idea of using the csv library and mock a file class:
```py
import csv
class fakefile:
    data = ""
    def write(self, data):
        self.data += data

def rows_to_csv(rows):
    file = fakefile()
    writer = csv.writer(file)
    writer.writerows(rows)
    return file.data
```

krestenlaust (approved):
Nice work!

I'm satisfied with this for now, but I've created these two issues #455 and #456, I would be greatful if you could take a gander at one of them at some point :)

PR #459, opened 2024-05-18
[Feature] Nix Flake

Mast3rwaf1z:
Added a nix flake, i have 3 things i'd like to test out with this, what is in this PR atm is just some things from #272 and some things from f-klubben/fnix#2

What i would like this flake to also do is to run the system by just calling `nix run github:f-klubben/stregsystemet` 

The devshell should actually already be usable by running `nix develop github:Mast3rwaf1z/stregsystemet`

- [x] Implement a devshell
- [x] Implement some way to ensure package version
- [x] Create default package


Mast3rwaf1z:
Still not done at all, the default package does not want to link the python packages properly, and as such the python path only works while working with the devshell

but to be fair, you wouldn't use a flake-based default package in a server side program without the devshell if its available? hmm

AshesOfEther (approved):

requirements.txt
krestenlaust, line 1: Is it neccesary to add these comments to the requirements? I think it might confuse/intimidate people only interesting in the python aspect of the project

krestenlaust (changes requested):

Mast3rwaf1z:
it *should* work still, without the comments :)

krestenlaust (approved):
Sorry for slow review! I'd mistaken it as unfinished.

LGTM! :shipit: 

You can merge at will!

Mast3rwaf1z:
> You can merge at will!

sure, let me just update the fork and test if it still works

Mast3rwaf1z:
i've updated the flake lock as well, so the newest nixpkgs can be used, everything still works so i'll merge

PR #464, opened 2024-05-21
Razzia overhaul (v2)

krestenlaust:
I'm trying to overhaul the razzia, but I'm more experience with django since I last tried.

I'll move the razzia into a separate app, and work on it from there.


krestenlaust:
#495 superceeds

PR #465, opened 2024-06-06
[Fix] Add remaining member data to userdata csv

Mast3rwaf1z:
@JakobTopholt pointed out that gender was missing in the userdata.csv file, i checked and noticed that more data was missing, so this PR will solve that.

### DIFF
--- a/stregsystem/mail.py
+++ b/stregsystem/mail.py
@@ -59,8 +59,8 @@ def send_userdata_mail(member):
     )
     userdata_csv = rows_to_csv(
         [
-            ["Id", "Name", "First name", "Last name", "Email", "Registration year"],
-            [member.id, member.username, member.firstname, member.lastname, member.email, member.year],
+            ["Id", "Name", "First name", "Last name", "Email", "Registration year", "Active", "Gender", "Want spam", "Balance", "Undo count", "Notes"],
+            [member.id, member.username, member.firstname, member.lastname, member.email, member.year, member.active, [text for (identifier,text) in member.GENDER_CHOICES if identifier == member.gender][0], member.want_spam, member.balance, member.undo_count, member.notes],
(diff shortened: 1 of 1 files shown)


stregsystem/mail.py
```diff
@@ -59,8 +64,34 @@ def send_userdata_mail(member):
...
+                member.firstname,
+                member.lastname,
+                member.email,
+                member.year,
+                member.active,
+                gender,
+                member.want_spam,
+                member.balance,
+                member.undo_count,
+                member.notes,
```
krestenlaust, line 93: Whether we want to have Notes, could be a topic for discussion, I've regarded it as an internal note about users used by Admins. On the other hand, maybe it legally is regarded as user data and we might as well include it 🤷‍♀️

It would be interesting to get a separate field of study field in the Member parameter and clear the notes

krestenlaust:
Unrelated but still interesting; to have categories as a column in purchases export

stregsystem/mail.py
```diff
@@ -59,8 +64,34 @@ def send_userdata_mail(member):
...
+                "First name",
+                "Last name",
+                "Email",
+                "Registration year",
+                "Active",
+                "Gender",
+                "Want spam",
+                "Balance",
+                "Undo count",
+                "Notes",
```
krestenlaust, line 79:
```suggestion
```

stregsystem/mail.py
krestenlaust, line 93:
```suggestion
```

krestenlaust: I've decided that we should just leave out notes until we find a good reason to include it 👌

krestenlaust (approved):
I haven't tested it, but the changes look reasonable 👍

PR #468, opened 2024-06-07
Changed username check to be case insensitive

### LINKED ISSUE #466: Admin panel username check is case-sensitive
This can result in an admin mistankenly creating duplicate usernames for members (since quickbuy is case-insensitive, one of them won't be able to access their accoun)

patrickreiffenstein:
Fixes #466.

### DIFF
--- a/stregsystem/admin.py
+++ b/stregsystem/admin.py
@@ -202,7 +202,7 @@ class MemberForm(forms.ModelForm):
     def clean_username(self):
         username = self.cleaned_data['username']
         if self.instance is None or self.instance.pk is None:
-            if Member.objects.filter(username=username).exists():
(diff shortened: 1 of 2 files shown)


krestenlaust (commented):
Thanks!

Can you take a look at #178 and why the tests added in that PR hasn't caught the issue we experienced here in #466 and update the tests accordingly?

krestenlaust (changes requested):
Thanks!

Can you take a look at #178 and why the tests added in that PR hasn't caught the issue we experienced here in #466 and update the tests accordingly?

krestenlaust (approved):
Awesome work! Thanks for doing the test

Issue #450, opened 2024-05-14
Setup guide for Mac users


krestenlaust:
Having a section for mac users would be nice

krestenlaust:
@many5900 @MadsLudvig

krestenlaust:
Fixed in #470

PR #470, opened 2024-07-03
Setup guide for Mac users. #450

BoAnd:
I think this should work :thinking: 

The the best solution would be they bought a lenovo :100:

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,19 @@ For Ubuntu with virtual envs:
 5. ???
 6. Profit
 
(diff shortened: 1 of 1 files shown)


krestenlaust (commented):
Awesome, Bo! I'll just reach out to some Mac users to have them try it out

krestenlaust:
I might set up an OSX virtual machine to test this at some point!

krestenlaust (approved):
Nvm, it's probably fine. Thanks!

Issue #474, opened 2024-08-26
Update deprecated <table> styling


atjn:
The `<table>` elements make heavy use of deprecated attributes.

It should be possible to replace these with CSS declarations.

The attributes in question:
- `align`
- `valign`
- `width`
- `cellpadding`
- `cellspacing`
- `border`

Issue #473, opened 2024-08-26
Sop using <center>


atjn:
The HTML documents use `<center>` extensively, but it is a non-standard and deprecated element.

It should be possible to replace it with a few CSS declarations.

PR #475, opened 2024-08-26
Upgrade products table to modern HTML

atjn:
I have removed the old table-based layout and used modern CSS instead. The visual change is basically nonexistent, you can switch between the old and new one without noticing it.

Most tables on the frontend use the same formatting, so I have designed a `default` style that I expect to use for most of the tables, in some upcoming PR's.

When you review, I recommend you hide whitespace diffs, because otherwise it seems like I have replaced the whole document, which I have not.

This is a partial implementation of #473 and #474.


stregsystem/templates/stregsystem/index.html
```diff
@@ -9,6 +9,18 @@
 {% endif %}
 {% endblock %}
 
+{% block head %}
```
jones-g, line 12: Shouldn't this be added to a css file instead of being placed in a head section of index?

atjn: That is one way of doing it, but it means I will need to create ~10 new small CSS files, one for each page that needs custom styling. I felt that it would be much simpler to have a small style block in the document where you can add these small tweaks. This is not just simpler for the developer, but also for the poor browser that already needs to download more than 5 separate stylesheets.

But if you think it should be separate, I can definitely do that :)

krestenlaust:
@atjn 
Most likely unrelated to this PR (but I discovered it just now)
#297 adds another button in the top, which causes the title to be off-center, do you have a good fix?
billede

krestenlaust (approved):
Looks good, I like the idea of adding a default style definition

PR #476, opened 2024-08-27
Remove prefixed declarations

atjn:
These prefixes are no longer necessary, they just clutter the files.

- animation declarations have been supported without prefix for 9 years: https://caniuse.com/css-animation
- keyframes have been supported without prefix for 9 years: https://caniuse.com/mdn-css_at-rules_keyframes
- user-select declarations have only needed the `-webkit-` prefix for 4 years: https://caniuse.com/mdn-css_properties_user-select


stregsystem/static/stregsystem/easter.css
```diff
@@ -67,144 +63,120 @@
...
     left: 42%;
-    -webkit-animation-delay: 7s, 1s;
     animation-delay: 5.69s, 1.66s
 }
 
 .beerflake:nth-of-type(23) {
     left: 69%;
-    -webkit-animation-delay: 2.5s, 1s;
     animation-delay: 42s, 1.27s
 }
```
krestenlaust, line 182: In some places like these, which have different values, which would be chosen by a browser?

Because if a browser assumes the webkit animation delay, it would need the values to be updated

krestenlaust (approved):
Nice to get some of the clutter out of the way👍

stregsystem/static/stregsystem/easter.css
jonasKjellerup: CSS always prioritizes the most significant (e.g. most specific & latest) supported property. So `-webkit-*` will only be used if the browser does not support the unprefixed version.

krestenlaust: Good to know!

PR #480, opened 2024-08-27
Upgrade header to modern HTML

### LINKED ISSUE #478: Header off-center because of buttons
Most likely unrelated to this PR (but I discovered it just now)
#297 adds another button in the top, which causes the title to be off-center, do you have a good fix?
billede

_Originally posted by @krestenlaust in https://github.com/f-klubben/stregsystemet/issues/475#issuecomment-2313268959_

atjn:
It turns out the heading text has never been centered. Here you can see:
 - 1. the original header
 - 2. the new header with a sign up button
 - 3. this new design which ensures that the heading text is actually centered

Skærmbillede fra 2024-08-28 00-42-34

And on a narrow screen:

Skærmbillede fra 2024-08-28 00-43-42

This new design also ensures that there is consistent spacing between the buttons and the edge of the screen.

Fixes #478.
Is a partial implementation of #473 and #474.

### DIFF
--- a/stregsystem/static/stregsystem/stregsystem.css
+++ b/stregsystem/static/stregsystem/stregsystem.css
@@ -3,6 +3,22 @@ body {
        background-color: white;
 }
 
+header {
+  display: grid;
+  grid-template-columns: 1fr auto 1fr;
+  padding: 3px 4vw;
+  column-gap: 4vw;
+  text-align: center;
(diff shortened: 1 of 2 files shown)


krestenlaust (approved):
Good work! Nice to have a styling wizard on the team🎉

PR #482, opened 2024-09-05
Add a theme system

atjn:
Partial implementation of #261

This is a standardized way to define and load themes. If a theme is not being shown at the current date, none of its code is delivered to the frontend.

This works by having a database table that defines each theme and the date range where it is shown. Django queries the table for current themes, and if any are found, it searches for the files to include in the frontend.

Having this in the database also means that themes can be enabled and disabled from the admin pane. I have called it "overrides", and it allows you to make beautiful and CPU-destroying abominations like this:

Screenshot of the stregsystem where all themes are enabled

For more information, I would recommend that you read `themes.md` where I document how to use the feature.

I was a little worried about the performance penalty of querying a database on each page load, so I have added a short 30 seconds cache to the database call. In a future PR, we might want to implement the proper Django caching system instead.

I have a lot of ideas on how this can be further improved and expanded, but the scope of this PR is already massive, so I plan on working on those ideas in future PR's.

I am a total Django noob so please review carefully, and any feedback and nitpicks are welcome, I would love to get better at it :)


AshesOfEther:
Keeping a database table just for something we already have in a JSON file is a bit overkill, no?

krestenlaust (approved):
`loadthemes` -> `reloadthemes`

Otherwise, nice work!

mads256h:
Why are we storing theme stuff in the database? A json file or something should suffice imo. Just my to cents tho :shrug:

atjn:
Have you tried reading the other comments?

PR #484, opened 2024-09-07
Hotfix: 'Ignore' status not being updated

krestenlaust:
When selecting 'ignore' as an option, the status didn't update correctly.

Tests should be added for this.


JakobTopholt:
Now ignoring a payment not set to a user gives this results, and ignoring a payment set to a user does nothing :(
image
Streg
@krestenlaust  :( Nice try tho.

PR #495, opened 2024-09-14
Refactor (foobar) razzia

krestenlaust:
The previous razzia code was very difficult to maintain, the goal is to combine Bread, Foobar and Fnugfald razzia into a single generic razzia. Eventually, the razzia wizard functionality can be implemented as well (but that's not a goal for this PR)

This builds on the experience from #464, but replaces it as well.


krestenlaust:
Foobar functionality is now implemented. The new razzia-system is hidden at `127.0.0.1:8000/admin/razzia`

The next step is to implement some way to change the razzia settings, but it is actually possible to do so using the new 'Razzia'-setting in the admin panel

atjn:
I did a basic review of this and everything seemed to work, but I don't exactly know how the whole razzia thing works, so I am afraid to formally review it :)

krestenlaust:
> I did a basic review of this and everything seemed to work, but I don't exactly know how the whole razzia thing works, so I am afraid to formally review it :)

Perfect timing. The basic purpose of the razzia system, is to validate that a member is actually a member. It also has the option for a timer between visits, based on how the Razzia object is defined

atjn:
I can approve it formally if you want me to, as long as you are aware that I am maybe not the best person to review it :)

I understand the purpose and use of the system, but I am unsure about how it works in the backend, and how it ties into the old razzia system. My lack of knowledge in that area means I probably missed some edge case tests and can't really give any thoughts on whether you designed the code correctly.

krestenlaust:
> I can approve it formally if you want me to, as long as you are aware that I am maybe not the best person to review it :)
> 
> I understand the purpose and use of the system, but I am unsure about how it works in the backend, and how it ties into the old razzia system. My lack of knowledge in that area means I probably missed some edge case tests and can't really give any thoughts on whether you designed the code correctly.

🙏  Normally I wouldn't want people to review if they aren't sure, but I think your experience is already the same as most other reviewers. Few FIT members actually know how the tools are used

atjn (approved):
Works on my machine™

PR #496, opened 2024-09-15
Improve GDPR implementation

atjn:
I made some improvements to the GDPR implementation:

- Instead of including it on every page, only include it on the one page that uses it.
- The scripts are now inline. External scripts might never load because of various network issues. Making them inline means that they either load, or none of the document is able to load.
- Also hide the text with CSS. The JS solution works fine, but it can break or be disabled for various reasons. Adding a complementary CSS solution makes it much more likely that the information is removed.
- No longer changes the length of the text. This means the layout won't jump around randomly 5 seconds after loading.
- Instead of replacing the text with a lot of "x"'es, just replace it with a single "_". This is mainly for assistive technologies; If you are using a screen reader, you will now hear "underscore just bought one beer" instead of "ex, ex, ex, ex, ex, ex, ex, ex just bought one beer".


atjn:
This is oviously not compatible with #196 but it would not be hard to implement that one on top of this PR. I would be happy to do the work on that one if need be :)

PR #498, opened 2024-09-19
Only load cursor styles when cursor is used

atjn:
This media query essentially says to only load the CSS when a classic mouse cursor is present. In all other cases there is no need to load the style, or it could even be detrimental because the input device could break when trying to apply this style.


krestenlaust (approved):
LGTM👍

PR #500, opened 2024-09-23
added icon

Mast3rwaf1z:
noticed that the stregsystem gets a 404 because there's no icon

ps. i made the branch on the wrong remote by mistake :/


krestenlaust (approved):
Nice👍

PR #502, opened 2024-09-24
Improve testdata

atjn:
A few improvements to the testdata. Most notably:

Added some extra users to easily test low balance features.

Added a lot of products to simulate the real stregsystem. This is especially important when doing UI work, since it needs to work with a large set of entries.

image


stregsystem/fixtures/testdata.json
```diff
@@ -1,111 +1,807 @@
...
+      "rooms": [1],
+      "categories": []
     }
   },
   {
     "model": "stregsystem.product",
-    "pk": 42,
+    "pk": 44,
     "fields": {
-      "name": "Fadøl, en is og et søm\"",
```
krestenlaust, line 65: Det var bevidst der var et citationstegn i. Det var for at teste CSV export funktionen

krestenlaust (commented):
Maybe we should just have a new fixture that closely resembles our current data. Falke and I discussed this issue somewhere on here, it would be great to have anonymized data or recreated data in same proportion

We should probably have a table discussing the choices for fixture and what they offer (and also clean em up a bit)

stregsystem/fixtures/testdata.json
atjn: Fair, my bad

atjn:
I agree with both ideas, but I can't access the database, so this is the best I can do. I think this would be a fine improvement that could then be followed by more improvements when you have time to add the anonymized data?

README.md
```diff
@@ -58,7 +58,13 @@ Admin panel: <http://127.0.0.1:8000/admin/>
...
 
 Stregsystem: <http://127.0.0.1:8000/1/>  
-User: `tester`
+There are different users that help you test different things:
+| Username | Superpower |
+|---|---|
+| tester | Default test profile |
+| q | Has short name for maximum testing speed |
+| nodough | Has no stregdollars |
+| lowdough | Only has 15 stregdollars |
```
krestenlaust, line 67: Might as well get the terminology right this time round!
```suggestion
There are different members that help you test different things:
| Member | Superpower |
|---|---|
| tester | Default test profile |
| q | Has short name for maximum testing speed |
| nodough | Has no stregdollars |
| lowdough | Only has 15 stregdollars |
```

krestenlaust (approved):
Looks good, nice work! You're welcome to merge (squash) when the comment has been resolved 👍

PR #506, opened 2024-09-26
Improve stregsystem icons

atjn:
This is an implementation of https://github.com/f-klubben/logo/pull/9 in the stregsystem.

This fixes a bunch of edge case behavior, most notably it fixes the logo colors when in dark mode. This is the existing setup:

image

And this is the new setup:

image


krestenlaust (approved):
LGTM!

PR #510, opened 2024-09-30
Improve QR Code Rendering

mads256h:
Instead of using the default scaling algorithm (bilinear?) use the pixelated image rendering algorithm so it is always nice and crisp :^)


krestenlaust (approved):
Good call! LGTM

PR #512, opened 2024-10-08
Add Activate Fklub Theme

soupglasses:
It seems someone has forgotten to pay to activate our F-klub account, so an effective reminder strategy has been added to help encourage that to change. Set to remind us every patch tuesday. :window: :smile: 

Screenshot from 2024-10-08 11-48-45

### DIFF
--- a/stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
+++ b/stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
@@ -3,10 +3,10 @@
     right: 0;
     bottom: 0;
     pointer-events: none;
-    font-size: 12pt;
+    font-size: 1em;
     color: rgba(45, 45, 45, 0.5);
     float: right;
-    padding: 36pt;
+    padding: 2em;
 }
 
 .activatefklub-container p {
@@ -15,6 +15,6 @@
 }
 
 .activatefklub-header {
-    font-size: 16pt;
+    font-size: 1.4em;
     font-weight: 600;
 }
--- a/stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
+++ b/stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
@@ -1,4 +1,4 @@
-.activatefklub-container {
+#activatefklub-container {
     position: fixed;
     right: 0;
     bottom: 0;
@@ -7,9 +7,14 @@
     color: rgba(45, 45, 45, 0.5);
     float: right;
     padding: 2em;
+    display: none;
 }
 
-.activatefklub-container p {
+.activatefklub-shown {
+    display: initial !important;
+}
+
+#activatefklub-container p {
     padding: 0;
     margin: 0;
 }
@@ -17,4 +22,10 @@
 .activatefklub-header {
     font-size: 1.4em;
     font-weight: 600;
+}
+
+@media (prefers-contrast: more) or (prefers-reduced-transparency) {
+    .activatefklub-shown {
+        display: none !important;
+    }
 }
--- a/stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
+++ b/stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
@@ -4,7 +4,7 @@
     bottom: 0;
     pointer-events: none;
     font-size: 12pt;
-    color: rgba(0, 0, 0, 0.33);
+    color: rgba(45, 45, 45, 0.5);
     float: right;
     padding: 36pt;
 }
--- a/stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
+++ b/stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
@@ -1,0 +1,20 @@
+.activatefklub-container {
+    position: fixed;
+    right: 0;
+    bottom: 0;
+    pointer-events: none;
+    font-size: 12pt;
+    color: rgba(0, 0, 0, 0.33);
+    float: right;
+    padding: 36pt;
+}
+
+.activatefklub-container p {
+    padding: 0;
+    margin: 0;
+}
+
+.activatefklub-header {
+    font-size: 16pt;
+    font-weight: 600;
+}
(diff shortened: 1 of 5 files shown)


soupglasses:
Ideas for alternative triggers:
- Run on April 1st only.
- Run on Stregsystem Remote only?

JakobTopholt:
I think it might run the issue that some people see their account as deactivated, perhaps it should have a slightly different text, such as "Look up Patch Tuesday for more info"

JakobTopholt:
Other than that, hilarious and I love it.

soupglasses:
Updated phrase to no longer reference an account to avoid such confusion.

image

stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
```diff
@@ -0,0 +1,20 @@
+.activatefklub-container {
+    position: fixed;
+    right: 0;
+    bottom: 0;
+    pointer-events: none;
+    font-size: 12pt;
```
atjn, line 6: Using `pt` is incredibly cursed. Use `em` instead, which adapts the font size to properly fit the device it will be displayed on.

stregsystem/static/stregsystem/themes/activatefklub/activatefklub.js
```diff
@@ -0,0 +1,20 @@
...
+attempt_spawn_activatefklub()
+
+function attempt_spawn_activatefklub() {
+    // Only activate on patch tuesdays.
+    if (new Date().getDay() == 2) {
+        spawn_activatefklub()
+    }
+}
+
+function spawn_activatefklub() {
```
atjn, line 11: This should not add HTML programatically, the HTML should be in the `.html` template and disabled with CSS `display: none`. The only thing this function should do, is

atjn:
That was not supposed to be individual comments, but here we are. They are just nitpicks, you don't have to fix them.

stregsystem/templates/stregsystem/themes/activatefklub/activatefklub.html
```diff
@@ -0,0 +1 @@
+<div class="activatefklub-container"></div>
```
atjn, line 1: This is a unique ID. It should be `id=`, not  `class=`.

stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
```diff
@@ -0,0 +1,20 @@
...
+    color: rgba(45, 45, 45, 0.5);
+    float: right;
+    padding: 36pt;
```
atjn, line 9: Also `em` here.

stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
```diff
@@ -0,0 +1,20 @@
...
+}
+
+.activatefklub-container p {
+    padding: 0;
+    margin: 0;
+}
+
+.activatefklub-header {
+    font-size: 16pt;
```
atjn, line 18: Also `em` here.

stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
```diff
@@ -0,0 +1,20 @@
+.activatefklub-container {
+    position: fixed;
+    right: 0;
+    bottom: 0;
+    pointer-events: none;
+    font-size: 12pt;
+    color: rgba(45, 45, 45, 0.5);
```
atjn, line 7: The text is pretty hard to read if you don't have perfect vision. If the user has [`prefers-contrast: more`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast) enabled, you might want to fix the contrast somehow. One solution is to just remove the box.

This is especially a problem on a small viewports where it interferes with the text on the screen:
image

atjn (approved):
Very nice theme!

I have some nitpicks here that you don't need to fix, but if you don't, I will probably fix them in a follow-up PR :)

stregsystem/static/stregsystem/themes/activatefklub/activatefklub.css
atjn: Same with [`prefers-reduced-transparency`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-transparency)

stregsystem/templates/stregsystem/themes/activatefklub/activatefklub.html
soupglasses: Still added the change, but might need to have a seperate PR for it for the other themes.

atjn (approved):
Beautiful, thanks!

stregsystem/themes.json
```diff
@@ -74,5 +74,17 @@
...
+        "name": "activatefklub",
+        "html": "activatefklub.html",
+        "css": "activatefklub.css",
+        "js": "activatefklub.js",
+        "begin": {
+            "month": 1
+        },
+        "end": {
+            "month": 12
+        }
```
krestenlaust, line 88: When is this theme to activate? We could have this theme just show some time after facking if you'd like? It's pretty fun

krestenlaust (commented):
When

stregsystem/themes.json
soupglasses: It shows every month on the 2nd tuesday of that month. So its happening "every month", but only for one of those days. Sadly there is no real way to encode that idea in the themes system so its implemented in the javascript instead. Its a very light theme (only some CSS and 2 HTML elements). Open to change the implementation to better fit the theme system as needed.

soupglasses: Added a comment to the json file to explain this. :+1:

JakobTopholt:
@soupglasses Merge?

krestenlaust:
:(

krestenlaust:
I did not review this

krestenlaust:
Squash :((
billede

soupglasses:
Oof :(

Maybe a good idea to disable "default merge" in project settings if it is not to be used?


image

PR #516, opened 2024-10-21
Improve bat theme

atjn:
Rewrite the bat theme to improve a number of things:

- The bats are now shown instantly on page load
- A lot of janky behavior is removed, e.g. all bats flapping at the same time instead of randomly
- The bats no longer require a [layout pass](https://web.dev/articles/rendering-performance#2_js_css_style_paint_composite) which makes them much less CPU intensive
- The theme mostly works without JS enabled
- If the CPU is struggling, the JS will automatically throttle itself to make sure the page is still responsive
- The code is now documented so you can figure out what is actually going on
- It now respects users that want reduced motion, reduced transparency, or increased contrast.
- If you hide the tab in your browser, the JS will stop trying to destroy your CPU


krestenlaust (commented):
The bats freeze on my pc. Works on your machine?

Recreate: `reloadthemes` -> open page -> frozen bats

atjn:
Are they completely frozen? Not even flapping? That sounds like you have `prefer-reduced-motion` enabled?

PR #519, opened 2024-10-24
Added constraint to signup form - No whitespace in username

pimber:
…rname. Django handles trailing and leading whitespaces it selfs, so that '-test10' all leadingwhite spaces is ingored by django when inserted. The same with 'test10-' all trailing is also ingored


pimber:
Note:
The regex handles the cases where there is a whitespace somewhere in the username like (- is whitespace):
"test-10" and "t-est-10" or any of the combinations in between.

The regex do not handle these cases: "-test10", "test10-" or "-test10-". But, these cases do Django handles itself and therefore any leading or trailing whitespace is ignored when inserted into the table. I tested it my self multiple time!

Alfa omega, whitespace is now not allowed :-)

krestenlaust:
Reverted earlier commit, to re-run black with project settings

krestenlaust (approved):
Nice work!

PR #522, opened 2024-10-30
Improve advent candle theme

atjn:
Rewrite the advent candle theme to improve a number of things:

- The SVG is significantly simplified (smaller file size, faster render, easier to read).
- Moved some CSS that was accidentally included in the base stylesheet.
- It now respects users that want reduced motion.
- Made the leaves look a bit nicer when the candle is fully burnt down.
- Works completely without JS!


krestenlaust (commented):
On my screen size, the candle gets yoinked down the page with the changes 🤔 
billede


Before:
billede

After:
billede
billede

atjn:
Did you call `reloadthemes`? Seems like you are just loading the old candle but the old CSS and JS is no longer there.

krestenlaust (approved):
Great work as usual 👍

PR #524, opened 2024-11-06
Better form id

atjn:
The original ID on the form is not very descriptive, this is just a tiny fix to make it better.

It was called "focusform" because the ID was only used in the hacky focus script which we have now removed.


krestenlaust (approved):
Nice!

PR #525, opened 2024-11-07
Add a footer

atjn:
Add a neat little footer.

This fixes the long standing issue that our TOS is not visible: #301

I also made this because some of our themes would be much easier to make if there was a proper footer to add content to. You will se the benefit in some future PRs :)

image

### DIFF
--- a/stregsystem/static/stregsystem/stregsystem.css
+++ b/stregsystem/static/stregsystem/stregsystem.css
@@ -1,6 +1,10 @@
 body {
-       color: black;
-       background-color: white;
+  color: black;
+  background-color: white;
+  margin: 0;
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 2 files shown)


krestenlaust:
Maybe #369 can be fixed here?

atjn:
@krestenlaust that sounds like a good idea, but it would require you to maintain the version number somewhere in the source code. Maybe this would be best for a separate PR where we could discuss how to set that up?

atjn:
I can only find a version number in [setup.py](https://github.com/f-klubben/stregsystemet/blob/next/setup.py) and it is not up to date. We need to figure out how to make sure it gets updated, then we need to figure out how to access it and add it to the page.

krestenlaust (approved):
LGTM

PR #526, opened 2024-11-07
Make header sticky

atjn:
Make the header stick to the top of the viewport so if you scroll down a web page, you can still use the header controls.

This works best when combined with #525 

image


krestenlaust:
Nice feature, but part of the stregsystem-charme 💯

PR #527, opened 2024-11-12
Improve menu form

atjn:
Improvements to the product buying form in the user menu:
 - No longer uses JS.
 - Page size is much smaller.
 - Form is much more accessible.

There are no visual changes, except for some edge cases.

Instead of creating a new form for each product, we create a single form with a custom submit button for each product. Because we don't have to create 1000 forms, we save a lot of bandwidth. This also contributes to accessibility, since the browser can assert that this is a single form with different options, rather than a thousand separate forms that might not be related to each other.

Instead of using a link that invokes JS to submit a form, we use a button that natively submits the form. This is a big accessibility improvement, as it is now clear that the button is related to the form, and it is clear that the product name in the button is highly relevant.

### DIFF
--- a/stregsystem/templates/stregsystem/menu.html
+++ b/stregsystem/templates/stregsystem/menu.html
@@ -12,8 +12,21 @@
     padding-top: .8em;
     padding-bottom: 1em;
   }
-  #productlist {
+  #product-buy-form {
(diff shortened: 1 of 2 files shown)


krestenlaust (approved):
Nice👍

PR #529, opened 2024-11-29
Razzia correct member stats

krestenlaust:
- Add unique member count to members page (New "fember(s)", the old is renamed to "entries")
- Correct term on menu page


krestenlaust:
Menu:
menu screen
Members:
members screen

AshesOfEther (approved):

PR #531, opened 2024-11-29
Razzia permissions

krestenlaust:
- Rename permissions and divide them
- Rename permissions to standard permissions

Add more detailed permissions for fine-grained control. This lets us create a limited user which can only let people enter their credentials, and now view member data and such


krestenlaust:
Now redirects to the latest razzia, when the user doesn't have permission to browse. Now razzia users can simply have the permission "razzia.view_razzia"

AshesOfEther (approved):

PR #443, opened 2024-05-01
Added next year to yearly report page

Zeps2369:
I added the next year button to the yearly report page. Next year will be hidden while the displayed year is the current year or more.

PR #532, opened 2024-11-30
stregreport: Add all categories to overview

### LINKED ISSUE #514: Admin panel ranking (Købs- og forbrugsrangeringer) not showing all categories
As you can see from the following screenshot:
image

It is not possible to view categories such as "Sodavand". The other categories should probably be added.

krestenlaust:
The current implementation is very crude and hardcodes the product IDs.
Instead we should use the categories, and make it dynamic.

Fixes #514

I've also renamed the stregreport menu-entries to more reasonable names


krestenlaust:
The fixture doesn't have any transactions, so it can't easily be tested (Without me having to go through great amounts of pain)
Screenshot of the new screen, where a bunch of categories are grouped tightly together

You could argue that the styling should be updated. But then I might argue that this admin page is already still 10x more useful now, than it ever has been (Of course, I don't mind, not having to mess around with styling)

stregreport/views.py
```diff
@@ -305,54 +305,19 @@ def sales_product(request, ids, from_time, to_time, error=None):
 def ranks_for_year(request, year):
     if year <= 1900 or year > 9999:
         return render(request, 'admin/stregsystem/report/error_ranksnotfound.html', locals())
```
Mast3rwaf1z, line 307: when you're redesigning stregreport anyway, isn't this kind of a useless check?

stregreport/views.py
```diff
@@ -305,54 +305,19 @@ def sales_product(request, ids, from_time, to_time, error=None):
...
-        1859,
-    ]
-    coffee = [32, 35, 36, 39]
-    vitamin = [1850, 1851, 1852, 1863, 1880]
 
     FORMAT = '%d/%m/%Y kl. %H:%M'
     last_year = year - 1
     next_year = year + 1
     from_time = fjule_party(year - 1)
     to_time = fjule_party(year)
```
Mast3rwaf1z, line 313: fjule_party, while being a funny name for a function, says very little about what the function does

stregreport/views.py
```diff
@@ -305,54 +305,19 @@ def sales_product(request, ids, from_time, to_time, error=None):
...
     kr_stat_list = sale_money_rank(from_time, to_time)
-    beer_stat_list = sale_product_rank(beer, from_time, to_time)
-    caffeine_stat_list = sale_product_rank(caffeine, from_time, to_time)
-    milk_stat_list = sale_product_rank(milk, from_time, to_time)
-    coffee_stat_list = sale_product_rank(coffee, from_time, to_time)
-    vitamin_stat_list = sale_product_rank(vitamin, from_time, to_time)
+
+    stat_lists = []
+    for cat in Category.objects.all():
+        stat_lists.append((cat.name, sale_product_rank(get_product_ids_from_category(cat), from_time, to_time)))
```
Mast3rwaf1z, line 319: i would've probably done something like: 
```py
stat_lists = [(
    cat.name,
    sale_product_rank(get_product_ids_from_category(cat)),
    from_time,
    to_time
) for cat in Category.objects.all()]
```
i think its more readable

stregreport/views.py
```diff
@@ -305,54 +305,19 @@ def sales_product(request, ids, from_time, to_time, error=None):
...
-        1846,
-        1847,
-        1853,
-        1855,
-        1856,
-        1858,
-        1859,
-    ]
-    coffee = [32, 35, 36, 39]
-    vitamin = [1850, 1851, 1852, 1863, 1880]
```
Mast3rwaf1z, line 343: wtf

Mast3rwaf1z (changes requested):
there's a few small questions, descide for yourself whether you want to fix any of them, but i don't think there's anything wrong with the current code

good change :+1:

stregreport/views.py
krestenlaust: It's definitely the most cursed part of the stregsystem, now replaced. I'm also working on the new razzia, hoping to completely remove razzia from `stregreport`-module

stregreport/views.py
krestenlaust: Haven't touched the function. But might try to find a better name while I'm at it though

stregreport/views.py
krestenlaust: Solved by aac3ee0

stregreport/views.py
krestenlaust: Resolved by 9f484d2

stregreport/views.py
krestenlaust: The function is widely, I decided that adding type hints, and a real docstring is sufficient for the function to explain itself. A longer name would make adjacent code (adjacent to calls) more noisy to look at

If you disagree you are of course free to suggest further changes👌👌

Resolved by e3edd50

krestenlaust:
I've tried giving the report types new names to be more explanatory (I've also reordered them)

billede

billede

stregreport/views.py
```diff
@@ -303,56 +303,22 @@ def sales_product(request, ids, from_time, to_time, error=None):
...
-
     FORMAT = '%d/%m/%Y kl. %H:%M'
     last_year = year - 1
     next_year = year + 1
     from_time = fjule_party(year - 1)
```
patrickreiffenstein, line 309: Why is last year variable not used when it is defined right above?

patrickreiffenstein: and why is next_year never used :thinking:

patrickreiffenstein: This is the only weird stuff I could find and it is really minor. Is it anything you want to change @krestenlaust ? Otherwise I see nothing wrong with this being approved.

krestenlaust: It took me a while to remember, but it's actually because they are used in the template, all local variables are 'injected' into the template:
`return render(request, 'admin/stregsystem/report/ranks.html', locals())` <-- locals()

`next_year` was added in #443 to be used in the template. I've decided to move the variables to the bottom to more clearly show that they are used in the template. I decided `year - 1` is still better than using the variable.

Resolved by 5d85f44

patrickreiffenstein (approved):
Looks good to me

Mast3rwaf1z:
@krestenlaust  i've just tried running it before my last review and code now, it seems the categories doesn't show properly for me using the testdata?
before:
Screenshot_20250113_182847
after:
Screenshot_20250113_182703

same testdata: testdata.json

krestenlaust:
> @krestenlaust i've just tried running it before my last review and code now, it seems the categories doesn't show properly for me using the testdata?

After bisecting, it turns out it was the 'pythonic' implementation that was faulty. I ended up simply reverting it, because it (to me) is actually harder to understand than the current imperative. Let me know what you think

Mast3rwaf1z (approved):
lgtm, works on my machine :+1:

PR #534, opened 2024-12-02
Improve snow theme

atjn:
Rewrite the snow theme to improve a number of things:

- The snowflakes are now shown instantly on page load
- There is no longer a flash of unstyled content when loading the page.
- The theme now properly handles pages that are longer than the viewport.
- Instead of hard-coding random snowflake positions, they are dynamically applied for a new look every time you reload the page
- Works completely without JS!
- It now respects users that want reduced motion, reduced transparency, or increased contrast.

I would like to do a bit more work on this in the future, but I am currently too busy and these fixes are already very good on their own.


krestenlaust (approved):
LGTM :shipit: 

How do we make it snow even more? (I get that this is 1:1 with the old theme, it would be cool to see it snow more the further into december we get, or just snow a little more in general)

krestenlaust:
It is on prod now 👌

atjn:
> LGTM :shipit:
> 
> How do we make it snow even more? (I get that this is 1:1 with the old theme, it would be cool to see it snow more the further into december we get, or just snow a little more in general)

It does snow more the further we get into december, we are currently at peasant levels of snow 🎄

krestenlaust:
> > LGTM :shipit:
> > How do we make it snow even more? (I get that this is 1:1 with the old theme, it would be cool to see it snow more the further into december we get, or just snow a little more in general)
> 
> It does snow more the further we get into december, we are currently at peasant levels of snow 🎄

great to hear then😄

PR #535, opened 2024-12-02
Update fabfile with new deployment

krestenlaust:
I don't think it is necessary to reload apache2.service anymore 🤔


fabfile.py
```diff
@@ -4,12 +4,15 @@
 @task
 def deploy():
     with cd("/data/stregsystem"):
-        sudo("systemctl stop apache2.service")
+        sudo("systemctl stop stregsystem.socket")
+        sudo("systemctl stop stregsystem.service")
```
krestenlaust, line 8: Is it necessary to shutdown both of them?

fabfile.py
```diff
@@ -4,12 +4,15 @@
...
         with settings(sudo_user='stregsystem'):
             sudo("git pull --ff-only")
             with prefix("source /data/stregsystem/venv/bin/activate"):
                 sudo("pip install -rrequirements.txt")
+                sudo("pip install -rrequirements-prod.txt")
```
krestenlaust, line 13: Would the order of these ever be important?

fabfile.py
```diff
@@ -4,12 +4,15 @@
...
                 sudo("python manage.py collectstatic --noinput")
                 sudo("python manage.py migrate")
                 sudo("python manage.py reloadthemes database")
-        sudo("systemctl start apache2.service")
+        sudo("systemctl reload apache2.service")
+        sudo("systemctl start stregsystem.socket")
```
krestenlaust, line 18: Should I also start stregsystem.service again?

krestenlaust (commented):
I've tried not overcomplicating the file

fabfile.py
AshesOfEther: No, it should only be necessary to stop `stregsystem.socket`

fabfile.py
AshesOfEther: Only if one overrides the other, which they don't, and probably shouldn't in any case

fabfile.py
AshesOfEther: systemd should automatically start Stregsystemet upon the socket receiving traffic. In any case though, starting it immediately means that we find out right away if it fails to start

fabfile.py
```diff
@@ -4,12 +4,15 @@
...
         with settings(sudo_user='stregsystem'):
             sudo("git pull --ff-only")
             with prefix("source /data/stregsystem/venv/bin/activate"):
                 sudo("pip install -rrequirements.txt")
+                sudo("pip install -rrequirements-prod.txt")
                 sudo("python manage.py collectstatic --noinput")
                 sudo("python manage.py migrate")
                 sudo("python manage.py reloadthemes database")
-        sudo("systemctl start apache2.service")
+        sudo("systemctl reload apache2.service")
```
AshesOfEther, line 17: Touching Apache is unnecessary now, since Stregsystemet is in a different process. It should only be necessary to deal with Apache if connection details change, which is outside the scope of this deploy function.
```suggestion
```

fabfile.py
krestenlaust: Alright, thx

fabfile.py
krestenlaust: What happens to the existing stregsystem.service? Or is it only active during a request

fabfile.py
krestenlaust: That's a good point, let's keep it then!

fabfile.py
krestenlaust: Makes sense

fabfile.py
AshesOfEther: Typo, whooops. I meant to say that it should only be necessary to shut down *`stregsystem.service`*. To answer your question, it keeps running once started

krestenlaust: Hmm, what happens if apache2 tries to invoke stregsystem.socket during a deployment, then?

fabfile.py
```diff
@@ -4,12 +4,14 @@
 @task
 def deploy():
     with cd("/data/stregsystem"):
-        sudo("systemctl stop apache2.service")
+        sudo("systemctl stop stregsystem.socket")
+        sudo("systemctl stop stregsystem.service")
```
VirtualSatai, line 8: Are the stregsystem.socket and stregsystem.service systemd files located in the repository too?

krestenlaust: They actually aren't 😓 , do you have an idea for where to store them? Currently, they are just on the server

krestenlaust: Merry christmas also 🎅

AshesOfEther (approved):

fabfile.py
AshesOfEther: They're specific to our deployment, hence they're on the server

VirtualSatai: It would be nice to be able to reproduce the deployement of stregsystemet. But it's up to you all now :)

krestenlaust: @AshesOfEther we could make a dedicated prod folder, we already have requirements-prod.txt.

Preferably it would be in a different repo, but FIT has tried that in different capacities https://github.com/f-klubben/server-config and https://github.com/f-klubben/fnix. Otherwise, we could make a dedicated repo for storing files related to the stregsystem deployment. Finally, we could also just add it to server-config to have it somewhere in-case ITS and the server goes awry at the same time

Mast3rwaf1z: i would prefer putting it into another repo
this is where fnix was a good idea, because it would ensure apache, systemd, and whatever else is on git

Dockerfile
Mast3rwaf1z, line 1: what's the point of this dockerfile? xD

Issue #494, opened 2024-09-14
Rework tab-indexing and mnemonics


krestenlaust:
Tab-indexing on the frontpage is inconsistent and it's difficult to navigate the page, either remove the custom tab-indexes, or redo them to make them intuitive. The same goes for the mnemonics (access keys), it would be nice to have some good shortcuts.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey

atjn:
For tab indexes, I would just remove them. The autofocus attribute will ensure that "buy" is the first thing you can click and/or tab to, and as you say, if you want to go back to the buttons in the header, you will intuitively expect to tab backwards, but for some reason we have made it so you need to tab forwards.

PR #540, opened 2024-12-10
Get rid of tabindex

atjn:
Tabindexes change the expected flow of the page to make it less intuitive, we should get rid of them.

Partially ununfixes #494


krestenlaust (changes requested):
I partly disagree. Every other page than the frontpage should have tabindexes removed.

But I think the frontpage should be in this order: Quickbuy -> buy -> restart -> menu -> tilmeld. I feel like it's more intuitive for them to be in that order🤔 Even though it isn't standard web. My main argument for us to ignore the web in this sense, is the way it's accessed (through the stregkiosk) it doesn't feel like a webpage, it's closer to a kind of game.

atjn:
> I feel like it's more intuitive for them to be in that order🤔

The first time I used the stregsystem, I wanted to access the menu, so I pressed `Shift` + `Tab` because that moves you back up towards the header in almost all apps, but the stregsystem did something else which confused me, so now I just always use the mouse.

krestenlaust:
Should we maybe make a kind of design guideline? To point to, with discussions like these 🤔

atjn:
Sounds good to me.

krestenlaust (approved):
LGTM

We'll see how long this lasts (till someone makes a pr for the opposite again xd)

PR #541, opened 2024-12-10
Make popup qr codes for external links

atjn:
Adds a field to enable kiosk mode for each individual room.

If kiosk mode is not enabled for the room you are in, it no longer blurs people's names.

If kiosk mode is enabled, it replaces the links in the footer with qr code popups:
image

The mode should be enabled in most rooms but can be disabled in the `Strandvejen - Remote` room.


krestenlaust:
While it's kinda intuitive to have rooms opt-into this mode, it is kind of misunderstood imo. The stregsystem is only ever a stregsystem, it isn't meant to be opened on non-kiosk computers. I've actually considered to redirect all page loads from non-stregkiosk to fappen, but it isn't really mature enough on desktop imo

atjn:
> You should be careful to not repeat terminology, as stregsystem already has the app named 'kiosk' which is the screen in the Jægerstue

I see your point, but my counter point is that the screen in the jægerstue is not a kiosk and has incorrectly used the terminology for what I am adding here

atjn:
> While it's kinda intuitive to have rooms opt-into this mode, it is kind of misunderstood imo. The stregsystem is only ever a stregsystem, it isn't meant to be opened on non-kiosk computers. I've actually considered to redirect all page loads from non-stregkiosk to fappen, but it isn't really mature enough on desktop imo

I don't understand why it is a problem that people use the stregsystem on their own computers. But if you prefer to just use the "kiosk" solution on all pages and remove the opt-in, we can definitely do that.

krestenlaust:
> > While it's kinda intuitive to have rooms opt-into this mode, it is kind of misunderstood imo. The stregsystem is only ever a stregsystem, it isn't meant to be opened on non-kiosk computers. I've actually considered to redirect all page loads from non-stregkiosk to fappen, but it isn't really mature enough on desktop imo
> 
> I don't understand why it is a problem that people use the stregsystem on their own computers. But if you prefer to just use the "kiosk" solution on all pages and remove the opt-in, we can definitely do that.

My main issue is when people 'misuse' the stregsystem. Probably the most common case for refund is people accidentally refreshing the page with post content. It also adds a pressure to maintain the stregsystem as a website (extra work you've mainly done in your last ~15 PRs, but if you didn't do it, nobody else would). With the footer with links being one of the first features directed at the website use-case, which actually ended up disrupting the 'real' use-case.

Summarized, more bugs and intended behavior appear, because there are more user journeys. More user journeys also requires more features = more code = more difficult to maintain. Chat me up in the hallway if I'm missing something

atjn:
I understand that you want to keep the focus on the "kiosk mode" and dont want to maintain a separate mode. So I will remove the option to disable kiosk mode.

I also understand that the refresh issue is a big problem but I am working on some ideas to fix that, so I don't see that as a long-term issue.

I think that we can strike a good balance where we continue to make the stregsystem available remotely, but with limited support.

krestenlaust (approved):
Sorry for tardy review. Project...

Yes, I agree that we can look into having a good balance. That is mostly because, I also like the way that Fappen is superfluous now, and we don't *have* to maintain two systems, and can instead have our members rely on this single page.

I like this change, but having it be dynamic would've been much cooler, that way other stray links, like the one's on the error page, would also be caught (this doesn't have to be solved in this PR, you may merge when you feel like it!)

billede

atjn:
Alright, I will merge this and consider improvements in the future.

PR #544, opened 2024-12-28
hotfix: Signup Gender not properly assigned to Member

krestenlaust:
The gender entered during signup is never passed to the created member, and members are just registered with unknown gender.


patrickreiffenstein (approved):
LGTM, it appears to run properly now

PR #548, opened 2025-01-07
Added a confirmation dialog to prevent accidental purchase

Daniel21b:


JakobTopholt:
@Daniel21b Thanks, it's quite nice of you to try and contribute.

With that said, confirmation of purchasing isn't needed, it'll slow down the process, and generally speaking the users of the system make so few mistakes that it would overall be a negative. In the case that someone makes that mistake, they usually contact and admin and have them correct it.

Again, thanks for the contribution, you're living the FOSS life.

mads256h:
On the other hand, I think that a "refund" button would work nicely. So you can undo the latest purchase if there was a mistake. This seems to me to be the best of both worlds, without any compromises.

krestenlaust:
Hello Daniel, thank you so much for looking into our system. Unfortunately, this feature is not something we are looking into implementing for the time being.

If you're still interesting in contributing, you're welcome to take a look at the issues labelled "hacktoberfest":
https://github.com/f-klubben/stregsystemet/issues?q=is%3Aissue+is%3Aopen+label%3Ahacktoberfest

Daniel21b:
> On the other hand, I think that a "refund" button would work nicely. So you can undo the latest purchase if there was a mistake. This seems to me to be the best of both worlds, without any compromises.

Thank you I will implement this future

PR #553, opened 2025-01-14
[Nix support] Unflakify the nix flake and add utility

Mast3rwaf1z:
I've moved the package and shell to its own files and moved the dependencies.

I know its not standard to move the default package under nix-support, but i think the root of the stregsystem is very cluttered with files already, so i didn't want to add another file `default.nix` and `shell.nix` there.

For non-nix people, flakes are experimental features, which means i'd like to add backwards compatibility by providing a `default.nix` and `shell.nix` seperately from the `flake.nix`, so the non flake users can continue to use their `nix-build` and `nix-shell` :P

Additionally i've added a nixos module and added it as a vm as a package. I need this functionality to improve my kiosk project: https://github.com/Mast3rwaf1z/kiosk/pull/5


krestenlaust (approved):
Nix files mean nothing to me, but readme looks nice👍

PR #560, opened 2025-02-20
Make users quickbuy member search case insensitive

### LINKED ISSUE #479: Quickbuy is case-sensitive with unfinished members
billede

billede

ThomasBow:
Fixes #479


krestenlaust (commented):
Awesome, great work!

Can you add a test to prevent this from occurring in the future?

PR #561, opened 2025-02-20
Hotfix: Product notes 

patrickreiffenstein:
Fixed that problem notes stopped products from showing up. The fix was found by @atjn.

Added test that makes it so we catch it the next time.


krestenlaust (approved):
Awesome that you've added a test as well. Have you tested that it fails without the fix?

I'm hoping that @atjn can review it this time😅

patrickreiffenstein:
> Awesome that you've added a test as well. Have you tested that it fails without the fix?
> 
> I'm hoping that @atjn can review it this time😅

Yes, I tested that it gives an error in tests when the fix is not applied.

krestenlaust:
Looking at it now, the namedtuples are a bit messy, might redo later

PR #562, opened 2025-02-20
Make invalid input gdpr compliant

### LINKED ISSUE #557: GDPR-feature doesn't hide recent username when invalid input is given afterwards
In the feedback the system provides when the input is in the wrong format or the username doesn't exist, the previous username is shown in the output

ThomasBow:
Fixes #557

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ stregsystem.log
 .direnv
 .envrc
 result
+/.vscode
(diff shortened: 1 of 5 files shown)


krestenlaust:
After further discussion, we've agreed this change isn't relevant in regard to GDPR and in general

krestenlaust:
I am confused

krestenlaust:
Just in general

Issue #555, opened 2025-01-22
Formalize a Design Guideline


krestenlaust:
To make it easier for contributors to design features without as much back and forth, it would be neat to have a design guideline.

The guideline should describe
 - The charming personality of the stregsystem
 - The context of the system (e.g. trust, cashier-less, volunteer work) (to address #548, #286 and #345 )

What could also be covered
 - Best practices for developing features (e.g. be aware of what app the feature should reside in, removing excess migrations in PRs)
 - How products are expected to be populated (and used, e.g. headers for events)
 - Workflows for volunteers
 - The current deployment of the application on the server

PR #565, opened 2025-04-03
FIX: Easter theme now spans entire body height

VinceUB:
Fixed an issue where the background image for the easter theme only spans the view, which means that if the page is too long to fit into a single screen, the bottom bit stays unbackgrounded


VinceUB:
Note: I haven't actually tried building this to see if the code works. But I have tried changing the CSS manually with the Firefox devtools

krestenlaust (approved):
I think this is a good change. The old way it was borders on insanity, stretching is much better. Relevant for #555

Issue #362, opened 2023-10-13
Individual duration for each kiosk item


krestenlaust:
Hacktober: The kiosk system is located in it's own directory and is separate from the Stregsystem.

Similar to #131

PR #566, opened 2025-04-03
Add custom kiosk duration times for each item

### LINKED ISSUE #363: Make kiosk slideshow interval customizable in admin panel
Currently, if I remember correctly, the interval is hardcoded in the javascript of the kiosk.

Hacktober: the kiosk is a separate system from the Stregsystem, but the admin panel is shared.

patrickreiffenstein:
This fixes #363 and #362.


krestenlaust (changes requested):
Looks good, and I've tested that it still displays like normally.

You should remake the migrations, so there's only a single file

ThomasBow (changes requested):
It is buggy on low time intervals, tested 100 ms and 2000 ms

krestenlaust:
Interesting fix...

ThomasBow (approved):
ShameBellGIF

PR #583, opened 2025-09-18
Add website support to kiosk

krestenlaust:
Kiosk now supports showing websites

Yes you can nest it too


ThomasBow (approved):
In general I think this lgtm, you could consider adding some testing data, so that future work wouldn't have to create their own items for testing.

PR #584, opened 2025-09-22
Make attaching debugger easier

ThomasBow:
Adds support for attaching a debugger given a sepcific argument, and outlines the detail in a debugger.md file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ stregsystem.log
 .envrc
 result
 *.qcow2
+
+# vscode user and workspace settings
+.vscode/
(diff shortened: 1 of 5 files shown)


patrickreiffenstein (commented):
I am not sure about the casing with --waitForDebug. It seems weird, but idk.
Maybe add some text about vscode setup where you could mention that the django environment needs to be selected, for us people who usually run with cmd. Maybe also add a reminder to actually download a python debugger to the person reading the guide. Other than that, it appears to work.

patrickreiffenstein (approved):
lgtm

krestenlaust (approved):
LGTM

PR #589, opened 2025-10-08
Reduce buy queries

### LINKED ISSUE #357: Poor performance on high amount of products bought (even with multibuy)
Right now, using multibuy doesn't utilize the fact that it's a single order, and instead treats every single item as a new purchase input with individual database lookups etc.

An example: A kind of trivial order of 100 items is extremely slow.

patrickreiffenstein:
This is the changes from my forked branch, which broke a bit. This resolves #357.


ThomasBow (approved):
lgtm

PR #599, opened 2025-11-06
gdpr when input is invalid

OmSa120224:
Resolves issue #557


krestenlaust:
:(

krestenlaust:
Try look into these issues: https://github.com/f-klubben/stregsystemet/issues?q=state%3Aopen%20label%3Ahacktoberfest

There's also the log spam bug, which is high priority

PR #600, opened 2025-11-07
Enter = Genstart

### LINKED ISSUE #515: Can't easily navigate back after wrong product with keyboard
billede
The above screen doesn't allow navigation using keyboard, you have to use the mouse to return

OmSa120224:
resolves #515

Pressing Enter is the same as clicking "Genstart" unless the cursor is in a input field. 
This also mean that if you are in your account and press enter, it will take you back to the home page.


ThomasBow (approved):
LGTM, nice and simple!

PR #617, opened 2026-03-01
[Auth] Member SSO Provider (OAuth2/OIDC)

### LINKED ISSUE #619: OIDC single-sign on flow
Add a way for external services to authenticate a fember and access their data. **This can also be used for a single unified login across f-club services for both volunteers and fembers.**

Example of SSO buttons:
Image

krestenlaust:
This PR adds
- Login form to authenticate members using their email (**One-Time Password member login**)
- Endpoints and logic needed for external OpenID login (**External SSO provider**)

**How does it authenticate a member?**
Whenever a Member logs in using the new login form (`/ffo/login`), a companion non-[staff](https://docs.djangoproject.com/en/6.0/ref/contrib/auth/#django.contrib.auth.models.User.is_staff) user is added, and a reference to the user is added to the member.

This user is then authenticated with requests, the same way django normally handles user logins.

**How does it look on an external service?**
Example buttons:
image

**To-do:**
- [x] Add proper SSO button template for easy copying
- [x] Handle duplicate User usernames, when creating companion user
- [x] Create OTP sign-in (right now it just accepts without proof of email)
- [x] Rename to 'ffo'
- [x] Find a way to replace dependency of custom fork: https://github.com/krestenlaust/django-oauth-toolkit/tree/2.4.0-custom-login-url
- [x] Add OpenAPI definiton
- [x] Update fixtures
- [x] Clean-up / tidy-up code
- [x] Add '.well-known' redirect at '/' to '/o'
- [x] Collapse migrations
- [x] Send actual email with OTP
- [x] Add `groups`-claim and embed all user groups
- [x] Document creation of `oidc.key`, make sure you can start a dev env without it
- [ ] Clean-up / tidy-up code again (in the end)
- [x] Add revoke-ability on user profile in stregsystem
- [ ] Add tests to verify that admin logins still work independently

**Related:**
* Reported feature-need upstream: https://github.com/django-oauth/django-oauth-toolkit/issues/1663 ~~(currently solved in [fork](https://github.com/krestenlaust/django-oauth-toolkit/tree/2.4.0-custom-login-url))~~ Solved using custom login view


krestenlaust:
Current 'ffo/login', stage 1
image

stage 2
image

krestenlaust:
Remember to add these checks to login
```
    if not member.signup_due_paid:
        return False

    if not member.signup_approved():
        return False
```

krestenlaust:
Extracted all login details into separate PR to make it more clean. This one will have to wait for that to be merged into main: #637

PR #606, opened 2026-01-07
Tickets

ThomasBow:
Prerequisite / depends-on: #620 
Prerequisite / depends-on: #632

PR #620, opened 2026-03-02
Add refund state to sales (to refund as admin)

krestenlaust:
cherry-picked from #606

related: #215


krestenlaust:
Bruh, I believe migrating fixtures accumulates them inside the next one

Edit: I solved migration

patrickreiffenstein (changes requested):
When opening a sale in the admin panel, it looks like you can fill out the refund information for a sale and apply it, but nothing actually happens. So either disallow that or make it actually work.

Refunded sales also still count towards statistics for the user. Such as ranking, user info and heatmap. There is also no way of telling whether a product have been refunded from the users point of view. So the purchase history should probably get a new column.

krestenlaust:
@ThomasBow :point_right: :point_left: what is the bad man saying to us?

ThomasBow:
What he is saying is sounds reasonable... sadly

patrickreiffenstein:
Womp womp

stregsystem/templates/stregsystem/menu_userinfo.html
patrickreiffenstein, line 1: It would be nice if you made the column dynamic so it doesn't need to show whether or not something has been refunded, if there is nothing refunded anyway.

stregsystem/models.py
patrickreiffenstein, line 1: It is very easy to raise the already refunded runtime error as an admin. Could you maybe just stop people from trying to refund something that is refunded instead of just throwing an error?

patrickreiffenstein (changes requested):

stregsystem/templates/stregsystem/menu_userinfo.html
ThomasBow: So basically the functionality you want is:
- If nothing has been refunded then the column should not show in the admin panel
- Is this dependent on what is being showed on the current page, as if the current page has nothing which has been refunded it should disappear too?

stregsystem/models.py
ThomasBow: I will refrain from using exceptions, I did not realize these had a different meaning when in production mode.

stregsystem/templates/stregsystem/menu_userinfo.html
patrickreiffenstein: I was actually mostly thinking about the user page

ThomasBow: But the functionality is as described above?

patrickreiffenstein: If nothing has been refunded, then the user should not be shown a column. Doing it for the admin page also makes sense

PR #621, opened 2026-03-03
Add fixture migration command

### LINKED ISSUE #634: Make django-admin tool/command to migrate all fixtures at once

krestenlaust:
Adds new command
`python manage.py migratefixture --fixture stregsystem/fixtures/testdata.json`

Also adds
`python manage.py migratefixture --update-targets`
Which updates the targets specified in `pyproject.toml`

And finally adds
`python manage.py migratefixture --all-fixtures` which does the first command for all fixtures


migrate-fixture.sh
```diff
@@ -5,17 +5,41 @@
 # This will load the specified fixture on current `next` branch,
 # then it switches to the specified branch, and migrates the database,
 # then it dumps the database to the same file.
-### 
+### To migrate all at once
+# BRANCH=refactor/date-attribute
+#
+# for f in stregsystem/fixtures/*.json; do
+#   sh migrate-fixture2.sh "$BRANCH" "$f"
```
ThomasBow: migrate-fixture2.sh does not exist?

migrate-fixture.sh
```diff
@@ -5,17 +5,41 @@
 # This will load the specified fixture on current `next` branch,
 # then it switches to the specified branch, and migrates the database,
 # then it dumps the database to the same file.
-### 
+### To migrate all at once
```
ThomasBow: Maybe explain how to actually use this?

ThomasBow (changes requested):
See comments, but to summerize maybe just move the all functionallity to a new file and then have it be migrate-all-fixtures and itstead of calling migrate-fixture2 it calls migrate-fixture?

migrate-fixture.sh
krestenlaust: true, I forgot

migrate-fixture.sh
krestenlaust: I will look into it

krestenlaust:
Original shell script:
https://github.com/f-klubben/stregsystemet/pull/621/changes/35277d88c199a79794dcd8294db8801667a75842

migrate-fixture.sh
krestenlaust: The new script should be more self-explanatory

stregsystem/management/commands/migratefixture.py
```diff
@@ -0,0 +1,102 @@
...
+import os
+
+
+class Command(BaseCommand):
+    help = "Migrate fixtures stored as /**/fixtures/*.json"
+
+    def add_arguments(self, parser):
+        parser.add_argument("--fixture", required=False)
+        parser.add_argument("--update-targets", action="store_true", default=False)
+
```
ThomasBow, line 16: Take a walk with me for a second, and imagine if there was a --all argument, instead of having to run the migrate-fixtures.sh script, which I believe was the whole idea behind this command initially ❤️

ThomasBow (changes requested):
Image

ThomasBow:
@krestenlaust check changes

stregsystem/management/commands/migratefixture.py
krestenlaust: Sounds good. I initially decided against it because there's no logical way to select amongst fixtures (because there's no standard way to discover them)

krestenlaust:
I haven't tested it, but the code LGTM :+1:  @ThomasBow

ThomasBow (approved):
I feel like this is kinda untestable, unless we would create a fake database and migrations, and fixtures and test if they can be updated, so lets test it in practice :)

stregsystem/management/commands/migratefixture.py
```diff
@@ -0,0 +1,115 @@
...
+import tempfile
+import os
+
+
+class Command(BaseCommand):
+    help = "Migrate fixtures stored as /**/fixtures/*.json"
+
+    def add_arguments(self, parser):
+        parser.add_argument("--fixture", required=False)
+        parser.add_argument("--all", action="store_true", default=False)
```
krestenlaust:
```suggestion
        parser.add_argument("--all-fixtures", action="store_true", default=False)
```

krestenlaust: Let's make it explicit since it's a long switch anyway

krestenlaust (commented):
I've forgotten how update-targets works. We should also consider documenting this new feature (but let's do that later)

stregsystem/management/commands/migratefixture.py
krestenlaust: Discussed in person

PR #626, opened 2026-03-04
addded escape key functionallity to log out

mantarias:


patrickreiffenstein:
Wouldn't it be better if we just switched from the enter button to escape, instead of having both? If we have both then there still is the issue of confusion of whether you buy or reset.

DeeKahy:
The thinking was that some people are already used to pressing enter, so why mess up their existing muscle memory. Do you want enter removed before this gets merged?

patrickreiffenstein:
Yeah, I am not approving it before Enter gets yeeted.

patrickreiffenstein (approved):
lgtm

PR #631, opened 2026-03-18
Update README with Arch Linux installation steps

mantarias:
Added installation instructions for Arch Linux users.


krestenlaust (approved):
Going forward we should consider looking at this vertically, and collect installations steps for different OSes at the same step. But this is fine for now

PR #633, opened 2026-03-18
Make product note products nullable

### LINKED ISSUE #632: Product note support for having no products
### Summary
`ProductNote` should support being created without any associated products.

### Motivation
Currently, a `ProductNote` requires at least one product to be linked. By allowing product notes with no products, the ticket system can use them to indicate whether a ticket will be purchased as a standby ticket — without needing to tie the note to a specific product.

### Proposed Behaviour
- Allow creation of a `ProductNote` with an empty product list (or no products at all)
- This should not break existing behaviour for product notes that do have products

patrickreiffenstein:
Fixes #632


ThomasBow (approved):
Nice great job doing the migrations :)
LGTM

PR #639, opened 2026-04-08
Preview env

EmilVorre:


Dockerfile.migrate
```diff
@@ -0,0 +1,11 @@
...
+
+WORKDIR /app
+
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY . .
+
+CMD sh -c "printf '[database]\nengine=django.db.backends.postgresql\nname=${POSTGRES_DB}\nuser=${POSTGRES_USER}\npassword=${POSTGRES_PASSWORD}\nhost=${POSTGRES_HOST}\nport=5432\n' > local.cfg && \
+    python manage.py migrate --noinput"
```
krestenlaust, line 11: Will I ever need to change this? Because it looks a bit complicated as is right here

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r requirements.txt gunicorn
+
+COPY . .
+
+# Write local.cfg from env vars at startup
+CMD sh -c "printf '[database]\nengine=django.db.backends.postgresql\nname=${POSTGRES_DB}\nuser=${POSTGRES_USER}\npassword=${POSTGRES_PASSWORD}\nhost=${POSTGRES_HOST}\nport=5432\n' > local.cfg && \
+    python manage.py collectstatic --noinput && \
+    gunicorn stregsystem.wsgi:application --bind 0.0.0.0:8080"
```
krestenlaust, line 13: I think there could be a more robust way to write a local.cfg 🤔

krestenlaust (commented):
Still an awesome project you've got going

PR #640, opened 2026-04-09
Maintenance: Update nix dependencies and modernize nix code

Mast3rwaf1z:
Hi, i was looking through your backlog of pull requests and the recent changes. I noticed that nobody touched the nix setup since i last used it to test the system. That's fine, but i would like to still have a look at the system sometimes so i'd like to also update the nix code occasionally.

I wanted to set the nix channel to unstable, but unfortunately django is broken on unstable, so you get stable nix.

Other than the nix updates, i have also adjusted a few files such that the system is still the exact same, but pyproject is able to generate an executable for the stregsystem.

For reviewer:
* my experience with django outside of stregsystemet is none, so if anyone more experienced can answer whether there is a better way to run manage.py from pyproject i would love to adjust it
* How is the system run in production? Preferably i want the nix build to output a production ready package to have something nice and testable
* Since unstable is not working, i fear that new versions of django will break the system. This is just for info, but i think you should be aware that this dependency might break in the future.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,16 +2,16 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1732749044,
-        "narHash": "sha256-T38FQOg0BV5M8FN1712fovzNakSOENEYs+CSkg31C9Y=",
+        "lastModified": 1775595990,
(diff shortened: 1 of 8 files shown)


krestenlaust (approved):
Looks Good To Us

PR #642, opened 2026-04-16
Bugfix: get_member_id only active accounts

Nicolai-Holmgaard:
Fixed issue of inactive accounts still getting found with `get_member_id` function.
This caused cases with users with duplicate usernames and the api throwing an error.


Nicolai-Holmgaard:
I love copilot 🫠

krestenlaust (approved):
This points to a bigger issue that's general across the system, you are welcome to look into it

PR #643, opened 2026-04-16
Bugfix: Razzia now queries for a single user, not a list

Nicolai-Holmgaard:
It used to query a list of users with the name, active or inactive. Now it only queries for a single active user, hence better.

This is a important fix for me, since i have a duplicate username with an old inactive account. This will fix my issues 😁


krestenlaust (approved):
LGTM :shipit:

PR #644, opened 2026-04-16
fix start date not respected

### LINKED ISSUE #296: Admin/Sale: Startdate not respected
when setting a start date in the future, this start date is not respected. I would expect that the start date would keep the product inactive and not show it in the sale window, until the start date, where it would be activated and shown,
image

patrickreiffenstein:
Fixes #296


krestenlaust (approved):
Looks Good To Us

PR #645, opened 2026-04-16
fix start date and quantity

### LINKED ISSUE #605: Product start date makes product unbuyable
When you make a new product and fill out the boxes. If you then add a start date to the product and keep the quantity 0, then the product will be unbuyable and you will get stregforbud error.

Could probably be solved together with #296.

patrickreiffenstein:
Now when quantity is 0, then it is because it is unlimited, as it should be. Fixes #605.


krestenlaust (approved):
LGTU :shipit:

PR #648, opened 2026-04-17
Remove Pillow/PIL dependecy

### LINKED ISSUE #647: Pillow dependency not nessesary
There are no parts of the codebase using pillow.

---

QR Codes are made with `qrcode` dep.

Searching for imports or usage of pillow (imported as `PIL`) yields nothing.

Running the service, and the frontend without the dependency in python 3.11 and 3.13 (with mysql db), no images (Rotating f-logo, qrcodes) yields no errors or glitches.

Naitsabot:
fixes #647

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 Django==4.1.13
-Pillow==10.3.0
 Coverage==7.5.1
 pytz==2024.1
 freezegun==1.5.1


krestenlaust (approved):
LGTM after discussion in issue thread

krestenlaust:
I suspect Pillow has been used in a field like this at some point

```
ERRORS:
stregsystem.Achievement.icon: (fields.E210) Cannot use ImageField because Pillow is not installed.
	HINT: Get Pillow at https://pypi.org/project/Pillow/ or run command "python -m pip install Pillow".

System check identified 1 issue (0 silenced).
```

krestenlaust:
This has been reintroduced in #573 which is just merged upstream

Good still a very good observation

PR #658, opened 2026-06-08
adds uv lock file

### LINKED ISSUE #657: UV for easy python package management

ThomasBow:


krestenlaust (approved):
Looks good to me. We will have to consider using only pyproject.toml instead of requirements.txt in the future. But let's not tackle that now

Issue #667, opened 2026-09-01
Make tilmelding button bigger in September


VinceUB:
This would make it more obvious how to melde yourself ind, and remind friends of nonfembers that they should also get meldt ind

krestenlaust:
@ThomasBow hotfix on server?
done

PR #671, opened 2026-09-02
Made the signup button bigger specifically in september

danielweper:
This implements #667, making the button very big in september

### DIFF
--- a/stregsystem/templates/stregsystem/base.html
+++ b/stregsystem/templates/stregsystem/base.html
@@ -38,7 +38,7 @@
   <div class="right">
     <input type="button" onclick="location.href='/{{room.id}}';" value="Menu"/>
     {% block signup_button %}
-      <input type="button" onclick="location.href='/signup'" value="Tilmeld"/>
+      <input type="button" onclick="location.href='/signup'" value="Tilmeld" {% if is_september %}style="font-size: 40pt;"{% else %}{% endif %}/>
     {% endblock %}
   </div>
 </header>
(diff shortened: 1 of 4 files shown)


krestenlaust (changes requested):
I think doing this javascript-based, or simply introducing a variable in the view function would be better
(I like the latter more. It should be quite simple)

Implementing a context processor is overkill for this niche usecase imo

danielweper:
> I think doing this javascript-based, or simply introducing a variable in the view function would be better (I like the latter more. It should be quite simple)
> 
> Implementing a context processor is overkill for this niche usecase imo

With the variable in the view function, would you then only do it for index? That is the original reason why I did it with the context processor, as it works in all views. It also works anywhere when using JS, but maybe it's better if it's only on index.

danielweper:
Since /index and /sale are distinct, it makes the most sense to have it everywhere, hence why it's JS instead of a variable in the /index view function

stregsystem/templates/stregsystem/base.html
```diff
@@ -78,6 +78,14 @@ <h1>
       restartBtn?.click(); 
     }
   });
+
+  // Big signup button in september
+  const today = new Date()
+  const isSeptember = (today.getMonth() === 8)
+  if (isSeptember) {
+    const signupBtn = document.getElementById("signupBtn");
```
krestenlaust, line 86: In theory this assignment is redundant. signupBtn is initialised the moment it is included in an "id" tag on an html element

krestenlaust (approved):
Nice and simple solution, I like it

You may merge when you feel like it

stregsystem/templates/stregsystem/base.html
danielweper: It seems to be semi-non-standard behavior, and is generally discouraged for readability reasons

krestenlaust: Cool, it was also more of a fun fact

PR #674, opened 2026-09-07
Remove duplicate tilmeldings mail call

ThomasBow:
Remove duplicate call as it gets called after being saved in signals.py

### DIFF
--- a/stregsystem/models.py
+++ b/stregsystem/models.py
@@ -185,7 +185,7 @@ class Member(BaseModel):  # id automatisk...
     username = models.CharField(max_length=16, validators=[no_whitespace_validator])
     year = models.CharField(max_length=4, default=get_current_year)  # Put the current year as default
     firstname = models.CharField(max_length=20)  # for 'firstname'
-    lastname = models.CharField(max_length=30)  # for 'lastname'
+    lastname = models.CharField(max_length=31)  # for 'lastname'
     gender = models.CharField(max_length=1, choices=GENDER_CHOICES)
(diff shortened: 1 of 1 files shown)


krestenlaust:
hmmmmmmmmmmmmmmmmm, i guess this could've been it all along. Was this introduced with my 'hotfix'?

stregsystem/models.py
```diff
@@ -185,7 +185,7 @@ class Member(BaseModel):  # id automatisk...
     username = models.CharField(max_length=16, validators=[no_whitespace_validator])
     year = models.CharField(max_length=4, default=get_current_year)  # Put the current year as default
     firstname = models.CharField(max_length=20)  # for 'firstname'
-    lastname = models.CharField(max_length=30)  # for 'lastname'
+    lastname = models.CharField(max_length=31)  # for 'lastname'
```
krestenlaust:
```suggestion
    lastname = models.CharField(max_length=30)  # for 'lastname'
```

krestenlaust (changes requested):
Did you discover anything? The bug has cropped up again

krestenlaust:
I haven't been able to reproduce the bug locally

ThomasBow:
I too have not been able to reproduce it locally, so i was thinking we just make it a field on the member, wether they've been sent one or not. Or alternatively we should add some logging and wait till it happens again. I did remove something which seemed like it would cause a double call, but i don't see how it isn't happening every time

